PR #44, opened 2020-04-13
Added the option "cursorhide"

wipeautcrafter:
- the option "cursorhide" was added


Jelmerro:
I have already researched the possibility of adding a "mousehide" setting, which did not work combined with "mouse" being enabled. This PR has the same issue as the implementation that I came up with (but did not commit as it was not working properly). For the same reason, I will not merge this PR for now.

I have created [an issue](https://github.com/Jelmerro/Vieb/issues/45) to document the requirements for such a feature. Please update this PR with a fix for the webview bug, or close the PR if you also don't have one. Finally, please follow the eslint rules that are included in the Vieb repository.

PR #75, opened 2020-10-17
add special path tests

rogerwim:


Jelmerro:
I changed the target branch to V3, as I made a workaround for a jest bug to make the tests work on that branch. I also want to know why there are two files with tests instead of one with all. Finally, please change the hard-coded path locations, these are bound to fail on other systems (and as such fail when I run them).

PR #76, opened 2020-10-31
Support forward/back mouse buttons.

gebner:
In electron, back/forward mouse button events are translated to ["app commands"](https://www.electronjs.org/docs/api/browser-window#event-app-command-windows-linux).

Maybe these should be mappable via `map` and co., but I wasn't sure whether that's the right approach.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -305,6 +305,10 @@ app.on("ready", () => {
     mainWindow.on("closed", () => {
         app.exit(0)
     })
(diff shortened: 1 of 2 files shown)


Jelmerro (approved):

Jelmerro:
Thanks for this, looks good to merge. We could think of having the mouse as part of the map command, but for now that's really keyboard focused. If you think there is added value in changing that, please open a new issue for it (or even better, make a new PR :wink:).

PR #77, opened 2020-10-31
Add incsearch option for incremental search.

gebner:


app/js/actions.js
```diff
@@ -298,6 +298,13 @@ const toggleFullscreen = () => {
     })
 }
 
+const incrSearch = () => {
```
Jelmerro, line 301: Can you add this action to the unSupportedActions list in `app/js/input.js` so that it can't be called as an action by the user? Currently when I do `:call action.incrSearch` I start a search based on the current url. Also, please keep the name consistent with the setting: "incSearch" instead of "incrSearch", or give it a more readable name with full words such as "incrementSearch".

app/js/settings.js
```diff
@@ -45,6 +45,7 @@ const defaultSettings = {
     "guitabbar": "always",
     "keeprecentlyclosed": true,
     "ignorecase": false,
+    "incsearch": false,
```
Jelmerro, line 48: Can we turn this on by default? What are your thoughts on this?

app/pages/help.html
```diff
@@ -534,6 +534,8 @@ <h3 id="keeprecentlyclosed">keeprecentlyclosed</h3>
         Toggle whether to keep recently closed tabs and allow them to be restored. This setting is unaffected by other history settings. Disabling this setting will result in the <a href="#action.reopenTab">action.reopenTab</a> being unable to restore previously closed tabs. If <a href="#restoretabs">restoretabs</a> is disabled, all open tabs will be moved to the list of recently closed tabs when quitting Vieb, unless this setting is also disabled (in which case they can not be restored).
         <h3 id="ignorecase">ignorecase</h3>
         Toggle for case-sensitive searching in the page. Any existing search will be updated to match this setting when either <a href="#action.nextSearchMatch">action.nextSearchMatch</a> or <a href="#action.previousSearchMatch">action.previousSearchMatch</a> are called.
+        <h3 id="incsearch">incsearch</h3>
+        Toggle for incremental search in the page.
```
Jelmerro, line 538: Can we include a bit more details about this feature for those not familiar with this Vim setting?

Jelmerro (changes requested):
Thanks for this PR, I left some comments to make it a bit better. Let me know what you think of them, especially regarding the default value of the setting.

Jelmerro:
Also, I forgot to check this for the previous PR, but you didn't update the changelog file `CHANGELOG.md`. I'll add this myself for the current batch of PRs if needed, but I just wanted to let you know that such a file exists :wink:.

app/js/actions.js
gebner: Done and renamed to `incrementalSearch`.

app/js/settings.js
gebner: I'm a big fan and apparently it's also enabled by default in neovim, so I've changed the default to `true`.

app/pages/help.html
gebner: Done.

gebner:
Sorry, didn't see the changelog.  I've added a line there.

Jelmerro (approved):

PR #78, opened 2020-10-31
Add :restart command.

gebner:
This command is really useful during development.  (Note that `app.relaunch()` does not immediately restart the electron process, it only takes effect when the process exits later.)

### DIFF
--- a/app/js/command.js
+++ b/app/js/command.js
@@ -177,6 +177,11 @@ const quitall = () => {
     ipcRenderer.send("destroy-window")
 }
 
+const restart = () => {
+    ipcRenderer.send("relaunch")
+    quitall()
+}
+
 const openDevTools = (position = null, trailingArgs = false) => {
     if (trailingArgs) {
         UTIL.notify("The devtools command takes a single optional argument",
@@ -481,6 +486,7 @@ const commands = {
     "quitall": quitall,
     "devtools": openDevTools,
     "reload": reload,
+    "restart": restart,
     "v": () => openSpecialPage("version"),
     "version": () => openSpecialPage("version"),
     "h": help,
@@ -521,6 +527,7 @@ const noArgumentComands = [
     "qa",
     "quitall",
     "reload",
+    "restart",
     "v",
     "version",
     "history",
--- a/app/pages/help.html
+++ b/app/pages/help.html
@@ -123,7 +123,7 @@
         <ul>
             <li><span class="command-block">:reload</span> - Reload all viebrc commands</li>
         </ul>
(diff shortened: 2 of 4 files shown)


app/pages/help.html
```diff
@@ -131,6 +131,8 @@ <h3 id=":reload">:reload</h3>
             <li>Finally run the <span class="command-block">:reload</span> command to run all viebrc commands</li>
         </ul>
         Of course it's also possible to make a mapping or custom command to do this for you.
```
Jelmerro, line 133: I would add a sentence here that leads into the next command, that mentions the restart command. Because the reload command describes how to achieve reloading to startup state without an actual restart, it would be nice to mention that it's not hard to restart Vieb as there is a command to do so.

app/pages/help.html
```diff
@@ -131,6 +131,8 @@ <h3 id=":reload">:reload</h3>
...
+        <h3 id=":restart">:restart</h3>
+        This command will restart Vieb.  It is functionally equivalent to running the <a href="#:quitall">:quitall</a> command, and then starting Vieb again.
```
Jelmerro, line 135: I think the example section should be present for all commands, even if it's simple, because it shows just how simple the command is in a short overview: "`:restart` - Restart Vieb". I like the current description, no need for changes there.

app/js/command.js
```diff
@@ -452,6 +457,7 @@ const commands = {
     "quitall": quitall,
     "devtools": () => TABS.currentPage().openDevTools(),
     "reload": reload,
+    "restart": restart,
```
Jelmerro, line 460: Please add "restart" to the list of "noArgumentComands" just below. This is a list of commands that don't have any arguments, and it will throw a generic warning for these commands if they contain trailing arguments.

Jelmerro (changes requested):

Jelmerro:
I apologize for the conflict, but I want this to be included in the V3 branch to prepare for the 3.0.0 release.

app/js/command.js
gebner: Done.

app/pages/help.html
gebner: I've added a link to the `:restart` command in the middle instead.  That sentence already mentioned restarting Vieb, and it seems to fit better there.

gebner (commented):
I've rebased and also added a changelog entry.

app/pages/help.html
Jelmerro: Even better, thanks for fixing this.

Jelmerro (approved):

Jelmerro:
Of course there is a conflict now that the incremental search has been merged, if you resolve these we can merge the restart command, looks good :+1:

gebner:
Ok, rebased.

PR #79, opened 2020-11-01
Allow binding <S-Space>

gebner:


Jelmerro:
I don't think this is the right fix for this issue, seems like the condition should be checking `keyCode` instead of `e.key`. I have updated the V3 branch to include a fix, please check if we still need this PR, if not, please close it. Thanks again for reporting and fixing so many Vieb issues :slightly_smiling_face:.

gebner:
Works perfectly on master now.

> Thanks again for reporting and fixing so many Vieb issues.

Sure thing!  It's surprisingly easy to work on Vieb!

PR #82, opened 2020-11-02
Several changes for a nix package

gebner:
I want to package vieb for [nixos](https://nixos.org).

The `electron-builder` tool does not work on nixos, so I'm calling the electron executable directly with the installed vieb app directory as argument.  For some reason, the current check did not reliably support this usage.  I have hence changed this part to use the `process.defaultApp` flag as described here: https://github.com/electron/electron/issues/4690#issuecomment-217435222

Furthermore, applications are installed read-only (with mode 0444 and modification time set to Jan 1, 1970).  The current adblocking code uses `fs.copyFileSync`, which unhelpfully preserves this read-only permission.  If you start vieb again, it fails because it cannot overwrite the now read-only blocklists in the config directory.  I have also added the error message to the failure notification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ The releases of Vieb aim to follow [semantic versioning](https://semver.org).
 - Mouse back/forward buttons being ignored when the mouse setting is enabled
 - Modifiers being ignored when pressing certain named keys such as "Space"
 - Border of split pages moving the page slightly when switching (border is now always there but in gray)
+- Updating the adblocker files when Vieb is installed on a read-only file system
 
 ### Deprecated
 
--- a/app/index.js
+++ b/app/index.js
@@ -131,9 +131,13 @@ const useragent = () => session.defaultSession.getUserAgent()
     .replace(RegExp(`${app.getName()}/\\S* `), "")
 
 // Parse arguments
-let args = process.argv.slice(1)
-if (args[0] === "app" || args[0] === app.getAppPath()) {
-    args = args.slice(1)
+let args
+if (process.defaultApp) {
+    // argv is ["electron", "app", ...args]
+    args = process.argv.slice(2)
+} else {
+    // argv is ["vieb", ...args]
+    args = process.argv.slice(1)
 }
 const urls = []
 let enableDebugMode = false
--- a/app/js/sessions.js
+++ b/app/js/sessions.js
@@ -54,9 +54,12 @@ const enableAdblocker = () => {
         for (const name of Object.keys(defaultBlocklists)) {
             const list = path.join(__dirname, `../blocklists/${name}.txt`)
             try {
-                fs.copyFileSync(list, path.join(blocklistDir, `${name}.txt`))
+                // File is read and written separately to discard permission
+                const body = fs.readFileSync(list)
+                fs.writeFileSync(path.join(blocklistDir, `${name}.txt`), body)
             } catch (e) {
-                UTIL.notify(`Failed to copy ${name}`, "err")
+                const msg = e.message || e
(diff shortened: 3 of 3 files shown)


app/index.js
```diff
@@ -131,10 +131,9 @@ const useragent = () => session.defaultSession.getUserAgent()
     .replace(RegExp(`${app.getName()}/\\S* `), "")
 
 // Parse arguments
-let args = process.argv.slice(1)
-if (args[0] === "app" || args[0] === app.getAppPath()) {
-    args = args.slice(1)
-}
+// If process.defaultApp is true, then argv is ["electron", "app", ...args].
+// Otherwise, argv is ["vieb", ...args].
```
Jelmerro, line 135: Please refactor this code to work on regular systems, `process.defaultApp` is undefined there, which means that this calculation evaluates to `undefined + 1` which is actually `NaN` (instead of 1, like you probably assumed). This refactor also gives opportunity to better use variables to hopefully require less explicit comments (though please do add them back if they are required for understanding it).

app/js/sessions.js
```diff
@@ -54,9 +54,13 @@ const enableAdblocker = () => {
         for (const name of Object.keys(defaultBlocklists)) {
             const list = path.join(__dirname, `../blocklists/${name}.txt`)
             try {
-                fs.copyFileSync(list, path.join(blocklistDir, `${name}.txt`))
+                // The copyFileSync function would preserve permissions
+                // (such as read-only)
```
Jelmerro, line 58: I don't think referring to the copy function actually adds a lot in this comment. Maybe a better comment would be `File is read and written separately to discard permission`. That is more clear to me about what is currently happening instead of saying why another way isn't suited. What are your thoughts on this?

CHANGELOG.md
```diff
@@ -56,6 +56,7 @@ The releases of Vieb aim to follow [semantic versioning](https://semver.org).
 - Mouse back/forward buttons being ignored when the mouse setting is enabled
 - Modifiers being ignored when pressing certain named keys such as "Space"
 - Border of split pages moving the page slightly when switching (border is now always there but in gray)
+- Updating the adblocker files when installed on a read-only file system
```
Jelmerro, line 59: I am not really familiar with NixOS, so correct me if I'm wrong, but shouldn't it be `when installed FROM a read-only file system`? To me it seems kind of impossible to install TO a read-only file sytem, and the fixes in this PR make more sense if it's FROM a read-only file system.

Please do tell me which of these is true, so I can better review this PR. I'll also try to read a bit about NixOS myself.

Jelmerro (changes requested):

app/index.js
gebner: Oh wow, this was not a boolean?!?  I've changed it and tested it in an ubuntu vm.

app/js/sessions.js
gebner: Ok, changed the comment.  Both explain why the next two lines shouldn't be refactored into a call to `copyFileSync`.

CHANGELOG.md
gebner: Ok, so this wasn't clear enough.  I've rephrased it to "when *Vieb is* installed on a read-only file system".
The issue happens when the vieb executable (and the bundled blocklists) are read-only.

app/index.js
Jelmerro: It depends, it's true when it's there, but otherwise it's undefined.

CHANGELOG.md
Jelmerro: Ah, now I understand it a bit more. Thanks for the clarification.

Jelmerro (approved):

PR #84, opened 2020-11-04
Fix detection of built electron app.

gebner:
Follow-up to #82.  There was a second occurrence of `argv.slice()` that I missed in the previous PR.

(I also accidentally introduced some eslint warnings, sorry.  These are also fixed now.)


Jelmerro:
Glad to see you made a function for it, I should have done that earlier, thanks!

PR #85, opened 2020-11-04
Do not save favicons with error status code.

gebner:
Sine websites don't have favicons (last one I ran into was https://nyxt.atlas.engineer) and you get a 404 response for the favicon.  In this case Vieb used to save an empty file, resulting in a "broken image" icon for the favicon.

There's still something strange going on with that site because the spinner does not disappear.


Jelmerro:
This seems like a good fix, but what do you mean with the spinner not disappearing? Here it seems to work, but the site does load a lot of images, so it might take a while for the page to load. Can you confirm that this is also the case on your end?

gebner:
> Here it seems to work, but the site does load a lot of images, so it might take a while for the page to load. Can you confirm that this is also the case on your end?

You're right.  That seems to be a reason.  There's a 20-second gif on that site, and the spinner only goes away once the gif is finished.

PR #88, opened 2020-11-13
Allow vertical scrolling to move the tabs list horizontally

gebner:
In mouse mode, you can now scroll up/down on the tab list to move it left/right.  (Scrolling left/right also moves the tab list as before, but that's not a comfortable scrolling direction on many mice.)


Jelmerro:
This seems like a great addition, I'll have time to test this in a few hours (hopefully).

PR #89, opened 2020-11-13
Add yank and download actions in follow mode.

gebner:
I'm not convinced this is the best way to do this but I'd like a way to copy link urls to the clipboard.  And also to download files that vieb would otherwise open directly (PDFs, but that's another issue).

This PR adds two new variants of the follow mode.  A yank mode (initiated by `action.startFollowYank`) which copies the link once picked, and a download mode (initiated by `action.startFollowDownload`) which downloads the picked link.  The default keybindings (`;y` and `;d`) are taken from pentadactyl; I believe qutebrowser uses the same ones.

 - I realize that this stretches the name of the "follow" mode a bit.  Pentadactyl/etc. use the name "(extended) hint mode" instead for the follow mode.

 - There is no visual indication what will happen once you pick a link (click, open a new tab, yank to clipboard, download).  This was already a (minor) issue since you couldn't tell if follow would open a new tab or not.  But now it's even more unclear as there are more actions that can take place after follow mode.

 - The download confirmation popup still shows up.  It might be a good idea to disable it in this case (and also in `pointer.downloadImage`).

### DIFF
--- a/app/pages/help.html
+++ b/app/pages/help.html
@@ -45,6 +45,9 @@
             <li><kbd>e</kbd> - Switch to <a class="explore" href="#action.toExploreMode">explore mode</a> (for navigating to websites or searching the web)</li>
             <li><kbd>f</kbd> - Switch to <a class="follow" href="#action.startFollowCurrentTab">follow mode</a> (for clicking buttons or links based on keys)</li>
             <li><kbd>/</kbd> - Switch to <a class="search" href="#action.toSearchMode">search mode</a> (for locating text in the page, use <kbd>n</kbd> and <kbd>N</kbd> to switch between matches)</li>
+            <li><kbd>v</kbd> - Switch to <a class="pointer" href="#pointer">pointer mode</a> (for hovering over links, downloading images, copying link urls, etc.)<ul>
+                <li><kbd>v</kbd> - Then switch to <a class="visual" href="#pointer.startVisualSelect">visual mode</a> (for selecting text)</li>
+            </ul></li>
         </ul>
         You can see the current mode at all times in the navbar, next to the url.
         Vieb is separated into different modes, most importantly normal mode.
@@ -804,6 +807,8 @@
         Distribute the windows inside a split evenly. This does not necessarily mean that all windows will be of the same size, for example if you have a vertical split, with on the left side a window, but on the right side a horizontal sub-split with two windows.
         <h2 id="pointer">Pointer</h2>
         Vieb has a <span class="pointer">pointer mode</span> to hover over links, download images or inspect an element at a position. The pointer can be moved in many different ways, most of which are by default mapped similar to Vim's normal mode. Additionally, after entering <span class="pointer">pointer mode</span>, it's also possible to start selecting text using <span class="visual">visual mode</span>. By default, the bindings of these two modes are similar, with the biggest difference being that text will be selected between the initial and current pointer position. See <a href="#pointer.startVisualSelect">pointer.startVisualSelect</a> to activate <span class="visual">visual mode</span> and <a href="#pointer.copyAndStop">pointer.copyAndStop</a> to copy the hovered link or selected text.
+        <h3 id="pointer.start">pointer.start</h3>
+        Switch to <a href="#pointer" class="pointer">pointer mode</a>, which enables a movable pointer.  The pointer can be used to hover over links, download images or inspect an element at a position. It can be moved in many different ways, most of which are by default mapped similar to Vim's normal mode. See <a href="#pointer.startVisualSelect">pointer.startVisualSelect</a> to activate <a href="#visual" class="visual">visual mode</a> and <a href="#pointer.copyAndStop">pointer.copyAndStop</a> to copy the hovered link or selected text.
         <h3 id="pointer.leftClick" countable="true">pointer.leftClick</h3>
         Simulate a left mouse click at the current pointer location. See also <a href="#pointer.rightClick">pointer.rightClick</a>.
         <h3 id="pointer.rightClick" countable="true">pointer.rightClick</h3>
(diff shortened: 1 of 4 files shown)


Jelmerro:
I was planning to tell you that an action named `pointer.copyAndStop` already exists, but the docs are very unclear for this action now that I read them. This action can actually be used from both pointer and visual mode, and can be used to either copy the selected text (while in visual mode) or the highlighted link (while in pointer mode). You can also combine pointer and visual mode with the follow mode (this is actually in the docs already, see `:h action.startFollowCurrentTab`). Considering this existing functionality, what would be the added value of the yank action? I'll update the docs to be more clear about this functionality soon.

About the download action, I would personally prefer to implement this outside of follow mode, either as an action for pointer mode, or as a separate `:download` command that takes a url as an argument (which you can get with the existing copy action). What are your thoughts on this?

Jelmerro:
I have updated the docs, hopefully this helps to understand the copy action that is already there a bit better.

Jelmerro:
Just to note, I don't really see the purpose in copying the mappings of other Vim browsers by using a `;` prefix. I would much rather give this a different mapping so that it fits better in Vieb with regards to the existing mappings.

With regards to the download popup, I would like to know how it actually differs in use from the `:write` command for your workflow.

gebner:
> I was planning to tell you that an action named pointer.copyAndStop already exists

This is cool, but there's no way I would have guessed that from the docs.  Instead of `;yabcd` I can just type `vfabcdy`, which is exactly what I was looking for!  The yank action is obsolete then.

> About the download action, I would personally prefer to implement this outside of follow mode, either as an action for pointer mode, or as a separate :download command that takes a url as an argument (which you can get with the existing copy action).

There is already `pointer.downloadImage`, so I'd make a `pointer.download` action for consistency.

> With regards to the download popup, I would like to know how it actually differs in use from the :write command for your workflow.

The `:write` command does something different than downloading a file: it saves an html file plus a directory containing images, etc.  Also the html file is modified (at least a comment is added at the beginning).  It also changes the file name.

By contrast `:download`/`pointer.download` would save the file as-is and without changing the filename.

There's also some additional complications with videos/pdfs.  If you open a video (e.g. a bare `mp4` file) and call `:write`, then vieb will save an html save with a `<video>` tag.  At least it plays the video when you open it.  It's even worse with PDFs: you get an html file with an `<embed>` tag that doesn't even contain the url of the pdf.

Jelmerro:
That sounds pretty bad actually, we might need a download action/command to make this a bit better. Feel free to update the PR to take care of this using the pointer.download action you proposed.

Also, did you check the new version of the docs I pushed recently? Let me know if they explain the copy action properly.

gebner:
I have now added a `pointer.downloadLink` action.

> Also, did you check the new version of the docs I pushed recently? Let me know if they explain the copy action properly.

The explanation for the `pointer.copyAndStop` action is great now.  However I have to admit that I wasn't even aware of pointer mode.  I always thought that this is just a way to simulate a mouse using the keyboard ("thanks, but my keyboard can already do that").

I think I overlooked it because it is pretty hidden in the documentation.  There is a list of four modes at the beginning of the help file, therefore I thought that I only need to care about those four modes.  The word "pointer mode" only appears a few pages in.  The `pointer.start` action is even missing completely.

I have added the pointer and visual modes to the list in the introduction, and added documentation for the `pointer.start` action.  Tell me what you think about it.  I am not sure `v` is the "canonical" keybinding for the pointer mode.

I believe pointer mode is pretty unique to vieb.  It deserves more prominent documentation.

app/pages/help.html
```diff
@@ -45,6 +45,9 @@ <h1 id="intro">Intro</h1>
             <li><kbd>e</kbd> - Switch to <a class="explore" href="#action.toExploreMode">explore mode</a> (for navigating to websites or searching the web)</li>
             <li><kbd>f</kbd> - Switch to <a class="follow" href="#action.startFollowCurrentTab">follow mode</a> (for clicking buttons or links based on keys)</li>
             <li><kbd>/</kbd> - Switch to <a class="search" href="#action.toSearchMode">search mode</a> (for locating text in the page, use <kbd>n</kbd> and <kbd>N</kbd> to switch between matches)</li>
+            <li><kbd>v</kbd> - Switch to <a class="pointer" href="#pointer.start">pointer mode</a> (for hovering over links, downloading images, copying link urls, etc.)<ul>
```
Jelmerro, line 48: Can we link this to `#pointer` instead of the start action specifically?

app/pages/help.html
```diff
@@ -810,10 +813,14 @@ <h3 id="pointer.rightClick" countable="true">pointer.rightClick</h3>
...
         <h3 id="pointer.inspectElement">pointer.inspectElement</h3>
         Inspect an element at the current pointer position. This action will open a devtools window if the devtools are not opened yet. The <a href="#devtoolsposition">devtoolsposition</a> is not used when inspecting an element, as it will open a devtools window. If you want to open the devtools differently, please open the devtools with the <a href="#:devtools">:devtools</a> command first. The "inspectElement" action will use the existing devtools to inspect the element in.
         <h3 id="pointer.insertAtPosition">pointer.insertAtPosition</h3>
         Enter <a class="insert" href="#action.toInsertMode">insert mode</a> at the pointer position. This can be used to click and immediately set focus to the element at a position, similar to an actual mouse click. Contrary to <a href="#pointer.leftClick">pointer.leftClick</a>, this type of click will also work for poorly programmed dropdowns and/or custom menus on websites: the ones that open with a click, but close automatically when focus is lost. Some Google websites use these type of inconsistent menus.
+        <h3 id="pointer.start">pointer.start</h3>
```
Jelmerro, line 822: I think it makes sense to make this the first action listed in the `#pointer` paragraph.

Jelmerro (changes requested):
Thanks for this PR, it's really great.

app/pages/help.html
gebner: Sure, I just tried to be consistent with the other items.

app/pages/help.html
gebner: Ok.

Jelmerro (approved):

PR #93, opened 2020-11-24
Style changes to click-to-open on downloads.

gebner:
- Render the filename in the download list as a link.
 - Accept clicks on the title in the download list.
 - Add a tooltip in the notification list.

I'm not sure there's a good way to make this discoverable in the pop-up notifications, short of adding a "(click to open)" text or a "↗" arrow.  But that seems a bit too in-your-face.

### DIFF
--- a/app/colors/default.css
+++ b/app/colors/default.css
@@ -245,7 +245,7 @@ webview {pointer-events: none;position: fixed;display: none;}
 #cookiespage .remove {height: 1em;width: 1em;margin: auto 0;cursor: pointer;}
 /* downloads */
 #downloadspage .download {background: var(--special-page-element-bg);margin-bottom: 1em;border: .1em solid var(--special-page-element-border);}
-#downloadspage .title {margin: .4em;font-size: 1.5em;width: calc(100% - 4em);}
+#downloadspage .title {margin: .4em;font-size: 1.5em;width: calc(100% - 4em);cursor:pointer;}
 #downloadspage progress {-webkit-appearance: none;width: calc(100% - 1em);margin: .5em;}
 #downloadspage progress::-webkit-progress-value {background: var(--download-progress-fg);}
 #downloadspage progress::-webkit-progress-bar {background: var(--download-progress-bg);}
--- a/app/js/preloads/downloads.js
+++ b/app/js/preloads/downloads.js
@@ -95,8 +95,12 @@ const addDownload = (download, id) => {
     element.appendChild(togglePause)
     // Title
     const title = document.createElement("div")
+    title.title = "Click to open"
     title.textContent = download.name
     title.className = "title"
+    title.addEventListener("click", () => {
+        ipcRenderer.send("open-download", file.textContent)
+    })
     element.appendChild(title)
     // Progress
     const progress = document.createElement("progress")
@@ -138,6 +142,7 @@ const addDownload = (download, id) => {
     downloadUrl.textContent = decodeURIComponent(download.url)
     misc.appendChild(downloadUrl)
     const file = document.createElement("span")
+    file.title = "Click to open"
     file.className = "filelocation"
     file.textContent = download.file
     file.addEventListener("click", () => {
--- a/app/js/preloads/notifications.js
+++ b/app/js/preloads/notifications.js
@@ -27,6 +27,7 @@ window.addEventListener("DOMContentLoaded", () => {
             element.className = "notification"
             if (notification.click?.type === "download-success") {
                 element.classList.add("filelocation")
+                element.title = "Click to open"
                 element.addEventListener("click", () => {
                     ipcRenderer.send("open-download", notification.click.path)
                 })


app/colors/default.css
```diff
@@ -245,12 +245,11 @@ webview {pointer-events: none;position: fixed;display: none;}
 #cookiespage .remove {height: 1em;width: 1em;margin: auto 0;cursor: pointer;}
 /* downloads */
 #downloadspage .download {background: var(--special-page-element-bg);margin-bottom: 1em;border: .1em solid var(--special-page-element-border);}
-#downloadspage .title {margin: .4em;font-size: 1.5em;width: calc(100% - 4em);}
+#downloadspage .title {margin: .4em;font-size: 1.5em;width: calc(100% - 4em);cursor:pointer;}
```
Jelmerro, line 248: Where is this used? On both the notification and download page has a pointer cursor due to the filelocation class.

app/js/preloads/downloads.js
```diff
@@ -97,6 +97,9 @@ const addDownload = (download, id) => {
     const title = document.createElement("div")
     title.textContent = download.name
     title.className = "title"
+    title.addEventListener("click", () => {
+        ipcRenderer.send("open-download", download.file)
```
Jelmerro, line 101: If we're changed the click region from just the filename to something else I think it would make more sense to have the entire block work as a file open action, with the same hover title you set on the notification page. This would make it work the same as on that page as well, to keep it consistent.

app/js/preloads/downloads.js
```diff
@@ -137,11 +140,13 @@ const addDownload = (download, id) => {
     downloadUrl.href = download.url
     downloadUrl.textContent = decodeURIComponent(download.url)
     misc.appendChild(downloadUrl)
-    const file = document.createElement("span")
+    const file = document.createElement("a")
```
Jelmerro, line 143: I don't like misusing elements for what they are not made for, which in this case is done by prevent the default click action of a link and sending the open ipc. I think it's nice to either have the entire block clickable with a simple title explaining the action, or keep it as is (possibly with just a hover title on the filename).

Jelmerro (changes requested):
Feel free to discuss the implementation below.

app/colors/default.css
gebner: This style is for the download *title*, which does not have the filelocation class.

app/js/preloads/downloads.js
gebner: I'm not against this, but it might be a bit confusing: there is already the URL link and the trash button in the download div, where a click means something completely different.

app/js/preloads/downloads.js
gebner: The only difference to a link is that the url doesn't open in vieb, but in an external program.  Before, I found it confusing that it wasn't styled as a link (blue and underlined).  I can change it back to a span, but I'd really like to keep the link styling (or some other indication that the file name can be clicked).

app/colors/default.css
Jelmerro: Right, now it makes a bit more sense, thanks for clearing that up.

app/js/preloads/downloads.js
Jelmerro: I get what you're saying, and I agree that we can improve the click functionality. Though I think that making the title clickable along with adding the hover text is good enough.  If people expect to be able to open the file with a click, the title and pointer cursor should be all the feedback they need to do so, and if using follow mode they will see the onclick outline as well.

As mentioned I don't like it being a link if it doesn't behave as one, interval Vieb pages should work as predictable as possible combined with follow mode, and this doesn't do that, let aside the fact that it's semantically [discouraged](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#onclick_events) to use a link for this.

gebner: > Though I think that making the title clickable along with adding the hover text is good enough.

This line of the diff is about the filename three lines below the title.  Are you suggesting that we should remove the filename and only have the title be clickable?  I kind of like this to be honest, the download div is a bit overcrowded.  (In any case, I'll add a tooltip to the title as well.)

Jelmerro: I meant that both the filelocation and the title can have the hover text and be clickable, but that changing the filelocation to a link shouldn't be necessary for that.

Jelmerro: Although it is crowded, I don't like removing it, as it contains info that isn't presented elsewhere in the download: the save location (which can be very handy in case of duplicate filenames, or when having recently changed the downloadpath).

app/colors/default.css
Jelmerro: Also, only the title of completed downloads should be clickable (and thus have a pointer cursor).

gebner: Opening an in-progress download can be useful as well, e.g. videos can be played while they're still downloading.

app/js/preloads/downloads.js
gebner: Ok, I've reverted the change and added the hover text.

app/colors/default.css
Jelmerro: Yes, for in progress downloads this would be nice, but I think the filename is different while downloading, and not all files can be opened that way. Finally it should be disabled for failed downloads regardless.

app/js/preloads/downloads.js
Jelmerro: Great, but there is one other thing we need to do that I just realized. The reason I used the textContent in the click event handler is because the download element gets updated when new files are added, or previous ones are removed. By using `download.file` instead, there will be disconnect between the click action and the filename displayed. The same is actually the case for the title. We can either use the textContent for both the handlers, or change the update function to remove the old listener and add the correct one.

gebner: Oh, I've changed it to use `file.textContent` in both cases.  I'm not going to add code to update event handlers.  It's probably a better idea to replace the whole element if a download has changed (or just use react to minimize changes to the dom).

Jelmerro (approved):

PR #94, opened 2020-11-24
Workaround for blocking shell.openPath() on Linux.

gebner:
Unfortunately, electron's `shell.openPath()` function blocks on Linux: https://github.com/electron/electron/issues/26074  Vieb freezes until the background application is closed again.

This PR adds a workaround and just calls `xdg-open` directly on linux.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -31,6 +31,7 @@ const {
 } = require("electron")
 const fs = require("fs")
 const os = require("os")
+const {spawn} = require("child_process")
 const path = require("path")
 const isSvg = require("is-svg")
 const {ElectronBlocker} = require("@cliqz/adblocker-electron")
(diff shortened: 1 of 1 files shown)


Jelmerro:
I personally didn't encounter this issue, but I don't see any reason not to merge this, I would expect most if not all Linux distributions to have `xdg-open` installed.

gebner:
The `shell.openPath()` function also calls `xdg-open` internally.

Jelmerro:
@gebner Can you test if we still need this workaround? The electron issue seems fixed, but I can't test if it's fixed, as I didn't have the issue before either. Would be great if you could let me know the results.

tsoil:
@gebner In which version is this problem solved，There is also this problem in 11.4.3

Jelmerro:
This issue was fixed in the [Electron repo](https://github.com/electron/electron/issues/26074), that's why I removed this workaround in Vieb. I asked Gebner for confirmation as a double-check to know for sure if the issue was resolved, but if you still have this in 11.4.3 then it would seem it isn't. Given that you are the author if that Electron issue, I would recommend to comment there to let them know that this isn't fixed for you.

gebner:
Ah sorry, I completely lost track of this issue.  Yes, I can confirm that this is still a bug in electron.  They haven't changed anything in the openPath implementation.

Jelmer van Arnhem ***@***.***> schreef op 13 mei 2021 12:06:43 CEST:

Jelmerro:
I recommend you to let the Electron developers know in the linked issue that it isn't fixed, I'll add back the workaround in the next Vieb release.

PR #95, opened 2020-11-27
Improve follow performance

gebner:
On some pages, entering follow mode can take a second or more on my machine.  Zulip is a particularly bad example.  This is a publicly accessible page where entering follow mode takes a long time: https://www.ikea.com/nl/nl/campaigns/slimme-koopjes-betaalbare-producten-voor-elke-dag-pub9015c218  Just computing the follow links (the `sendFollowLinks` function) takes 492ms for me.

This PR adds some easy speed-ups, without changing the general approach; the result should still be the same.  With the change, the ikea example only takes 70ms now.

I've split the `sendFollowLinks` function into multiple parts to facilitate profiling.  Most of the time (75%) is spent in the `parseElement` function.

Can you explain what problem the change to the link detection in 5697f580f165bf6620517b734ef1de23882c9651 fixes?  It adds quite a bit of runtime (25% in the example).

### DIFF
--- a/app/js/preloads/follow.js
+++ b/app/js/preloads/follow.js
@@ -69,15 +69,16 @@ ipcRenderer.on("focus-first-text-input", async () => {
 
 const getLinkFollows = allLinks => {
     // A tags with href as the link, can be opened in new tab or current tab
-    allLinks.push(...allElementsBySelectors("url", urls))
-    document.querySelectorAll("a *").forEach(el => {
-        const baseLink = el.closest("a")
-        if (baseLink && !allLinks.includes(baseLink)) {
-            const clickableSubElement = parseElement(el, "url")
-            if (clickableSubElement) {
-                clickableSubElement.url = baseLink.href || ""
-                allLinks.push(clickableSubElement)
-            }
+    document.querySelectorAll(urls.join(",")).forEach(e => {
+        const baseLink = parseElement(e, "url")
+        if (baseLink) {
+            allLinks.push(baseLink)
+        } else {
+            // Try subelements instead, for example if the link is not
+            // visible or `display: none`, but a sub-element is absolutely
+            // positioned somewhere else.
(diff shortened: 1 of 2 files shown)


Jelmerro:
The commit you reference is to detect absolutely positioned sub-elements of links, where the base link isn't shown. So for example, there is a `a` element, which has display none, or some obscure other position, but a sub-element is absolutely positioned somewhere else. In which case, the sub-element should still be detected as a link.

app/js/preloads/follow.js
```diff
@@ -69,27 +67,26 @@ ipcRenderer.on("focus-first-text-input", async () => {
...
-    allElements.filter(el => el.matches("a *")).forEach(el => {
-        const clickableSubElement = parseElement(el, "url")
-        const baseLink = allElements.find(
-            l => l.nodeName === "A" && l.contains(el))
-        if (baseLink && !parseElement(baseLink) && clickableSubElement) {
-            clickableSubElement.url = baseLink.href || ""
-            allLinks.push(clickableSubElement)
+    document.querySelectorAll("a *").forEach(el => {
+        const baseLink = el.closest("a")
+        if (baseLink && !allLinks.includes(baseLink)) {
```
Jelmerro, line 75: This check isn't correct and never matches, `allLinks` is a list of parsed links, while `baseLink` is a DOM element. Please double check with the diff from the code on master. You can test this on the newtab page, there should only be 1 follow element per link.

Jelmerro (changes requested):
I really like these changed, glad that you managed to improve the speed. There is just a single thing that isn't correct compared to version on master.

app/js/preloads/follow.js
gebner: Ah, what was I thinking! I've also added the explanation from your other comment.

Jelmerro (approved):

PR #97, opened 2020-12-01
Make addEventListener/removeEventListener wrapper more robust.

gebner:
This fixes an error with the follow mode on https://observablehq.com/@eric-wieser/lean-ga-imports

On that page, `WeakSet.add` unexpectedly throws an exception when adding a `HTMLButtonElement`.  This is probably a bug in chrome, because you should be able to add any non-null object.

I haven't run across this bug anywhere else, and it only happens for a single button on that page.  So this PR just ignores the exception.


Jelmerro:
Though I can't reproduce the bug you mention on that page, the new code is easier to read and is a lot simpler.

PR #98, opened 2020-12-03
Optimize history suggestion performance.

gebner:
- Only fetch favicons for the first suggestexplore suggestions.
 - Cache the computed simpleUrl

This reduces the runtime of suggestExplore() from 200ms to 16ms (cold: 46ms) for me.

### DIFF
--- a/app/js/history.js
+++ b/app/js/history.js
@@ -22,6 +22,7 @@ const path = require("path")
 
 const histFile = path.join(UTIL.appData(), "hist")
 let groupedHistory = {}
+const simpleUrls = {}
 let histWriteTimeout = null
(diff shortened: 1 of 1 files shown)


gebner:
The "fix for the suggestions mouse hover bug" from 6de5b71eb8ef942762c201fcc7286de8991cc57d now takes 10% of the runtime of suggestExplore.  Can you explain what the bug was?

Jelmerro:
This is really cool, I'll test this in a bit. It makes a lot of sense from just the code.

Regarding the "hover bug", I think that was just that the hover had an effect where you don't expect it if you have mouse disabled.

Jelmerro:
Works perfect, and fast too, thanks!

Issue #90, opened 2020-11-14
Feature Request: Theming support


B4rc1:
I would really like to switch to vieb from firefox + tridactyl,
the only thing missing is theming support like firefoxses userChrome.css.
I know this would be possible through forking the project and editing the css files, but I'd rather not maintain that for the rest of my life. Is there the possibility for a mechanism like a userChrome.css, which allows to inject custom css into the Vieb?

Jelmerro:
I think it would be cool to have this, but ideally we should strive for a design that works great out of the box for most people. If there are things you don't like about it I would much rather have a setting for it. Custom themes should be used only for custom themes, not to work around missing features, visual annoyances or other things that deserve more attention and polishing within Vieb.

From a technical view, I think it's possible to add this. I propose a `user.css` at the root of the datafolder to add custom styling to Vieb, which will be applied at startup (and also when running the restart command).

Finally, I usually don't tag people, but I think it would be really valuable to have more people involved in this discussion, especially because @gebner has made a wide range of CSS changes both as PRs and in their own fork of Vieb. I think these should either be integrated in Vieb, or be available as a custom theme, without having to keep a fork up to date with styling patches (though feel free to do so, as long as you inform people of the changes made in some way, as required by the [GPL3 license](https://github.com/Jelmerro/Vieb/blob/master/LICENSE)).

B4rc1:
I totally agree, everything should *just work* and there should be no requirement for a `user.css`.

Additionally there should also be a way to access the "browser internal" devtools, attached to the webview of the main UI (tabbar, navbar, ...) for theming experimentation and debugging purposes.

Jelmerro:
That's already there, use `--debug` as a startup argument. Do you have any example of things you would want to change with a `user.css` file?

B4rc1:
a lot actually, i would try to replicate my firefox-look which has:
1. tab and url bar at the bottom
2. change colorscheme (of everything) to to my main colorscheme (inspired by [iceberg.vim](https://github.com/cocopon/iceberg.vim))
3. remove the logo
4. (maybe depending on looks) move/style the mode indicator

the most important for me is the coloscheme change, because right now on my linux machine it sticks out.

Jelmerro:
[...]

PR #99, opened 2020-12-03
Add internaldevtools command to open developer tools of main window.

gebner:
It's very useful to be able to quickly open the devtools for a running vieb instance, e.g. to figure out why some action takes longer than expected.

I didn't think too hard about the command name.  It might be nice to open the internal dev tools as a tab, but this would probably require larger changes to addTab().

This already came up in another issue: https://github.com/Jelmerro/Vieb/issues/90#issuecomment-727245720 https://github.com/Jelmerro/Vieb/issues/90#issuecomment-727229717

### DIFF
--- a/app/pages/help.html
+++ b/app/pages/help.html
@@ -465,6 +465,11 @@
             <li><span class="command-block">:devtools tab</span> - Open the developer tools of the current page in a new tab</li>
         </ul>
         The devtools command can be used to open the developer tools of the current page. If no argument is given for the devtools position, it will open at the position configured by <a href="#devtoolsposition">devtoolsposition</a>. The accepted positions for this command are the same as that of the setting. It's not possible to open multiple instances of the devtools for the same page, you can open one type of devtools per tab. By default these tools will have dark mode and some security settings enabled. If the devtools are opened as a Vieb tab or split (so not windowed) all regular Vieb action can be used to interact with the developer tools.
+        <h3 id=":internaldevtools">:internaldevtools</h3>
+        <ul>
+            <li><span class="command-block">:internaldevtools</span> - Open the developer tools of the Vieb main browser window</li>
+        </ul>
+        The internaldevtools command opens the developer tools of the main Vieb window (containing the url bar, tab list, etc.).  This is mainly useful to debug Vieb itself.
         <h3 id=":colorscheme">:colorscheme</h3>
         <ul>
             <li><span class="command-block">:colorscheme</span> - Show the currently active colorscheme</li>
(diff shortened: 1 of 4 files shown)


app/pages/help.html
```diff
@@ -465,6 +465,11 @@ <h3 id=":devtools">:devtools</h3>
             <li><span class="command-block">:devtools tab</span> - Open the developer tools of the current page in a new tab</li>
         </ul>
         The devtools command can be used to open the developer tools of the current page. If no argument is given for the devtools position, it will open at the position configured by <a href="#devtoolsposition">devtoolsposition</a>. The accepted positions for this command are the same as that of the setting. It's not possible to open multiple instances of the devtools for the same page, you can open one type of devtools per tab. By default these tools will have dark mode and some security settings enabled. If the devtools are opened as a Vieb tab or split (so not windowed) all regular Vieb action can be used to interact with the developer tools.
+        <h3 id=":internaldevtools">:internaldevtools</h3>
+        <ul>
+            <li><span class="command-block">:devtools</span> - Open the developer tools of the Vieb main browser window</li>
```
Jelmerro, line 470: The example command is incorrect.

Jelmerro:
I expect that this was only tested with mouse enabled. Because if I use the `--debug` option, I can close the devtools again with `Ctrl-Shift-I`, but that doesn't work when using this command (without `--debug`). I am not too familiar with the devtools shortcuts myself, so I would expect Vieb's help docs to include a way to close them. Alternatively, there could be an option to switch the focus between the devtools and the Vieb. Or finally, we could force the devtools to open in a new window instead of in this embedded position (not sure if that's default or not, but that will not work correctly if not opened with `--debug`). This last option could actually make `--debug` simpler as well.

We could specify the [position to open the tools via the "mode" property](https://www.electronjs.org/docs/api/web-contents#contentsopendevtoolsoptions). I would suggest either "detach" or "undocked". With undocked we keep the option for users to move it back to within the Vieb window. Though there seems to be no way to close them using the keyboard if moved back to the same window (you either need to use the mouse in that case, or undock the tools again). We could actually change the default postition for the `--debug` tools to "undocked" as well, so we won't need the window frame at all.

In short, I want to keep using the keyboard to open or close the devtools, and undocking them by default seems to be the simplest way to achieve this, without requiring the window frame to be there. Please let me know your opinion on this suggestion.

gebner:
I don't have a strong opinion on the devtools placement, so I've set the `undocked` mode now.  I didn't really notice this because electron will remember the choice if you leave the mode field empty (and the first thing I did when testing was to undock it).

Jelmerro (approved):

Issue #100, opened 2020-12-10
Bookmarks


zyv42:
There is no way to bookmark/save a visited page with Vieb.
It would be really nice if there was a way to just issue a command like "addbookmark" and then get that website as a top suggestion when you start typing its title in Explore mode.
Better yet, it would be really neat to have a separate page "bookmarks", like already existing "downloads" or "history" where a user would be able to manually add, edit and delete bookmarks(probably even organize them with tags/directories.

Currently, it seems that the only ways to have a set of bookmarks is to add a list of "favoritepages" or use a separate html file(like bookmarks exported from another browser), both being lacking and inconvenient.

Jelmerro:
What exactly do you think is inconvenient about the `favoritepages` setting? Because to me, it sounds fairly similar to what you explain.

zyv42:
Ok, I'll try to write in details what exactly is different between `favoritepages` and bookmarks I have in mind:
First of all, `favoritepages` are not suggested in explore mode, so currently they are more like Firefox's "pinned" websites on Newtab rather than bookmarks.
Second, the interaction flow. Like, every time I'd like to add a new favorite page, I'd need to either issue a command `set favoritepages=`, which rewrites `favoritepages` entirely, so not an option if you have a few of those, or I'd need to go to viebrc file and edit it manually and then reload the browser for changes to take effect. Same goes for removing a favorite page.

So, I believe, it would be of great avail to be able just to issue some command like `addfavorite` and `delfavorite` to add/remove a current page's address from the `favoritepages`. And, to make `favoritepages` being suggested as a prompt in explore mode, separately from `topsites`.

Or, probably, simply leave `favoritepages` as they are, and implement a separate category of page collection, called `bookmarks` which will be prompted in explore mode like topsites and commands and with separate commands like `addbm` and `delvm` and maybe a separate page where you'd be able to manage(add/edit/delete) those bookmarks, like already existing `downloads` or `cookies`. In absolutely best case scenario it would be nice to have some user-given keys/titles for those website addresses and tags to be able to organize/group bookmarks for cases when you have many bookmarks.

Just to clarify, when I say "inconvenient", I mostly come from experience with other similar browsers, and even most minimalist ones, like vimb and surf provide ability to a degree to save and delete bookmarks.

Jelmerro:
[...]

PR #111, opened 2020-12-18
Implementing "quickmarks"/marks feature

InconsolableCellist:
Implementing marks to address #100 

I still have to make the help page and make default bindings of `M` and `g` to visit the marks, and do some more testing.

What do you think about this approach, though? I added two commands, `mark` and `retrieveMark`, ~~which call similarly named actions,~~ added a default setting and validation, and added a UTIL to get the URL for the mark setting.

~~I tried to follow the existing flow but I think that I could just do all the logic in command.js, and not call `ACTIONS` at all. I'll try that next.~~ *Edit:* I've now implemented this in the latest commit

This isn't ready for check-in yet, but I wanted to get some feedback on this approach.

### DIFF
--- a/app/js/actions.js
+++ b/app/js/actions.js
@@ -380,6 +380,43 @@ const setFocusCorrectly = () => {
     }
 }
 
+const mark = (inputMark, inputUrl) => {
+    if (!UTIL.getUrlForMark(inputMark)) {
+        newMapping = ""
+        SETTINGS.get("marks").split(",").forEach(mapping => {
+            const [mark, url] = mapping.split("~")
+            if (mark && url && requestedMark
+                && mark === requestedMark) {
+                url = inputUrl;
+            }
+            newMapping = `${newmapping},${mark}~${url}`
+        })
+    } else {
+        let marks = SETTINGS.get("marks")
+        if (marks) {
+            marks = `${marks},${inputMark}~${inputUrl}`
+        } else {
+            marks = `${inputMark}~${inputUrl}`
+        }
+        SETTINGS.set("marks", marks)
+    }
+}
+
+const retrieveMark = (requestedMark) => {
+    let outputUrl = ""
+
+    SETTINGS.get("marks").split(",").forEach(mapping => {
+        const [mark, url] = mapping.split("~")
+        if (mark && url && requestedMark
+            && mark === requestedMark) {
+            outputUrl = url;
+        }
+    })
+
+    return outputUrl
+}
+
+
 module.exports = {
     toCommandMode,
     toExploreMode,
@@ -447,5 +484,7 @@ module.exports = {
     commandHistoryNext,
     toggleFullscreen,
     useEnteredData,
-    setFocusCorrectly
+    setFocusCorrectly,
+    mark,
+    retrieveMark,
 }
--- a/app/js/actions.js
+++ b/app/js/actions.js
@@ -349,7 +349,7 @@ const useEnteredData = () => {
             if (searchword && url && query
                 && location.substr(0, searchword.length) === searchword
                 && /\S/.test(query)) {
-                    location = UTIL.stringToUrl(url.replace(/%s/g, query))
+                location = UTIL.stringToUrl(url.replace(/%s/g, query))
             }
         })
 
@@ -381,17 +381,21 @@ const setFocusCorrectly = () => {
 }
 
 const mark = (inputMark, inputUrl) => {
-    if (!UTIL.getUrlForMark(inputMark)) {
-        newMapping = ""
+    if (UTIL.getUrlForMark(inputMark)) {
+        // Update URL for an existing mark
+        let newMapping = ""
         SETTINGS.get("marks").split(",").forEach(mapping => {
-            const [mark, url] = mapping.split("~")
-            if (mark && url && requestedMark
-                && mark === requestedMark) {
-                url = inputUrl;
+            const [savedMark] = mapping.split("~")
+            let [, url] = mapping.split("~")
+            if (savedMark && url && inputMark
+                && savedMark === inputMark) {
+                url = inputUrl
             }
-            newMapping = `${newmapping},${mark}~${url}`
+            newMapping = `${newMapping},${savedMark}~${url}`
         })
+        SETTINGS.set("marks", newMapping)
     } else {
+        // Add a new mark
         let marks = SETTINGS.get("marks")
         if (marks) {
             marks = `${marks},${inputMark}~${inputUrl}`
@@ -402,14 +406,14 @@ const mark = (inputMark, inputUrl) => {
     }
 }
 
-const retrieveMark = (requestedMark) => {
+const retrieveMark = requestedMark => {
     let outputUrl = ""
 
     SETTINGS.get("marks").split(",").forEach(mapping => {
-        const [mark, url] = mapping.split("~")
-        if (mark && url && requestedMark
-            && mark === requestedMark) {
-            outputUrl = url;
+        const [savedMark, url] = mapping.split("~")
+        if (savedMark && url && requestedMark
+            && savedMark === requestedMark) {
+            outputUrl = url
         }
     })
 
@@ -486,5 +490,5 @@ module.exports = {
     useEnteredData,
     setFocusCorrectly,
     mark,
-    retrieveMark,
+    retrieveMark
 }
--- a/app/js/command.js
+++ b/app/js/command.js
@@ -526,7 +526,7 @@ const callAction = (...args) => {
     }
 }
 
-const mark = (markToSet) => {
+const mark = markToSet => {
     if (!markToSet || !markToSet.trim()) {
         UTIL.notify(
             "Exactly one mark character is required for the mark command",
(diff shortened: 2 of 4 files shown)


Jelmerro:
I'm not a fan of this. I don't see value in copying this over from other plugins if it doesn't fit in with how other things are done in Vieb. There are already favoritepages, which can be opened with a similar amount of keystrokes, without requiring you to remember which letter you assigned to which "quickmark". The added value of bookmarks from what I understood from the [parent ticket](https://github.com/Jelmerro/Vieb/issues/100) is that bookmarks should be MORE visual to the user, which is opposite of these "quickmarks", that hide the list of bookmarks in some sort of setting.

I'm going to repeat part of my comment from #100 here, but if bookmarks were to be added to Vieb, they should add something that favoritepages cannot do. If you ask me, that should be making them more visual with a special bookmarks page, where you can also tag/group/color them for easy reference, much more similar to regular browsers. These "quickmarks" don't really do that, as the only differences with favoritepages are the lack of a visual list and a different way to open them.

InconsolableCellist:
The purpose of quickmarks is to act like marks in vim, where you can quickly set and recall something without having to do more complicated visual management. It addresses the user's request of a faster way to set/access bookmarks, but beyond that it's a popular vim-like feature. Vim, Tridactyl, Pentadactyl, Vimperator, Qutebrowser, etc. all have a concept of marks and recall for pages.

It's something I really miss from Pentadactyl, as I use it constantly. 

>without requiring you to remember which letter you assigned to which "quickmark".

You don't remember it, it's muscle memory, the way you don't remember "f is for follow mode," you just know to press it without thinking about it.

>There are already favoritepages, which can be opened with a similar amount of keystrokes

How does one open a favoritepage using a keystroke? Like the mneumonic "go to Slashdot in a new tab" which in other Vim-inspired browsers would be `gns`, with `s` bound to the Slashdot URL. `h` might be home, `w` weather, `a` Amazon, etc.

Jelmerro:
I think your request is different from the feature requested in #100. The aim of that clearly seems to be to make it more visual. The complaint regarding the convenience seemed to have more to do with the misunderstanding that you would have to redefine the favoritepages each time, but you could just append the list with the `+=` operator. The author of that issue didn't know about that obviously.

It's not actually that complicated to open a new tab with a favorite page. By default, just pressing `t` will show you the list, with `f` and the assigned letter being all keys needed to be pressed to have any favorite page open in a new tab. But if all you want to do with these "quickmarks" is to have a specific set of keys open a page, why not just make a mapping for it. That's exactly what custom mappings are for I would say. This PR to me just seems like a different way to do this:
`nmap gh tevieb.dev<Enter>`
If you wanted to make the newtab part optional, just leave out `t` and either press that first to open a newtab or execute the mapping as is for the current tab. I really don't see how this PR makes this process any easier.

This seems just like a copy of what other browsers are doing. While that doesn't have to be a problem, it feels out of place to introduce "quikmarks" in Vieb when there are already other ways to achieve something similar. This is amplified by the fact that this doesn't in any way address the concerns of visibility mentioned in the #100 ticket. Here's a quote from the first comment of that issue:
`it would be really neat to have a separate page "bookmarks", like already existing "downloads" or "history" where a user would be able to manually add, edit and delete bookmarks(probably even organize them with tags/directories.`
So even if we were to merge this, would a THIRD system for bookmarks eventually need to be added to actually solve #100? I don't think that's right, and as such I want us to properly think this whole "quickmarks" idea through in a bit more detail.

InconsolableCellist:
The user requested a bunch of visual things but you had already replied, so I offered a method that's pretty similar to Vim functionality, and what other Vim-like browser addons/browsers do. I don't think there'd be a third method of handling bookmarks, as you didn't want tht visual management method.

With the favoritepages feature you'd still have to map things rather than using marks with `M` or similar. I'd consider that a functional workaround, but still a workaround. It seems to come down to what features you're used to in a vim-like browser. Personally I like quickmarks

Jelmerro:
I'll quote again from the parent issue here:
`I don't particularly like to make changes to explore mode for this, but feel free to open a PR for implementing a bookmarks page. My only requirement for such a page is that it indeed works similar to other special pages, and that it isn't just a favoritepages clone. It should offer functionality beyond that, such as the tagging or directories you mentioned.`
I don't see how this advice can possibly lead to this PR and your conclusion. It doesn't address either of those points. There is no bookmarks page implemented, and besides some convenient keybindings this doesn't offer much beyond favorites. Basically:

- I don't see value in copying just the mark functionality as is
- A bookmark page should be added that offer new functionality compared to favoritepages (such as directories or coloring)

This is mostly what caused my previous extensive comment. Because if we merge this, we don't actually solve #100 at all. Which in turn would mean that we need YET ANOTHER bookmark-ish feature. I don't like that, 2 should be enough. So this PR either needs to properly implement a nice bookmark page, or it won't be merged. And that's not because I don't see value in making it easier for people to do these "quickmarks", but because I want to do implement bookmarks right the first time, without requiring yet another bookmark feature later down the line. I don't use bookmarks myself at all, for me favoritepages are enough. But clearly there is a group of people that want bookmarks, and some have a lot of them (6000+). My point with that, is that I don't see people using these quickmarks for to manage/organize that many bookmarks. And as such, it doesn't actually implement bookmarks properly. So it's not that this PR isn't good, it's just not the way this needs to be implemented at this time. This PR feels rushed to just get back the familiar `m` shortcut, without considering the other requirements and existing structure.

I think going at this from the perspective of the bookmark page is better, because that we can actually solve #100 while doing this. Let's first make this bookmark page, give it tags, coloring, directories and all that kind of convenience stuff, and then implement a command to make and load bookmarks, based on for example name or tag, with auto-complete. And if that is all ready, we can start worrying about implementing these convenience features to load and save them.

InconsolableCellist:
Like I said, the purpose of my PR for quickmarks is to address the ability to quickly set and retrieve bookmarks, which I saw as obviating the need for the visual page that was being shut down in #100, and thus sounded like it wasn't going to happen otherwise. To me, quickmarks solves that issue in other browsers. 

I can simply make another issue and link to it if that's a sticking point. But it sounds like if I did open a feature request for quickmarks, you wouldn't want it upstream. If that's the case be upfront about it and I won't bother continuing

Jelmerro:
Where do I "shut down" the idea of a visual page? I feel like my previous comment went unread, and if that's the case, there is indeed not much point in leaving this PR open.

InconsolableCellist:
>Where do I "shut down" the idea of a visual page?

InconsolableCellist:
I meant you shut down the user's idea he said here:

>So, I believe, it would be of great avail to be able just to issue some command like addfavorite and delfavorite to add/remove a current page's address from the favoritepages. And, to make favoritepages being suggested as a prompt in explore mode, separately from topsites.

...

>In absolutely best case scenario it would be nice to have some user-given keys/titles for those website addresses and tags to be able to organize/group bookmarks for cases when you have many bookmarks.

He wanted a faster way to quickly set/retrieve bookmarks. I don't see having a visual page or suggestions as addressing it. He was unaware of using `+=` with `:set` but that alone doesn't do it, so I suggested adding quickmarks to be able to quickly set/visit pages, like Pentadactyl, Vimperator, etc.

But I can make this simpler and don't really care if it links to #100 or some other issue number: Should I make an issue for adding quickmarks and link this PR to it instead? Or will you not permit that feature to be in your version of Vieb? It was on my list of things I wanted to add anyway, as I use it multiple times a day in Pentadactyl, which I'm trying to switch from. (Same reason I added keyword searches.)

Jelmerro:
Okay, so that confirms that we were clearly talking past each other with regards to this PR. I don't mind a quickmarks feature, but I assumed that they wanted a visual page as well, because of this:
`it would be really neat to have a separate page "bookmarks", like already existing "downloads" or "history" where a user would be able to manually add, edit and delete bookmarks(probably even organize them with tags/directories.`
As such, I want to make sure we implement a bookmarks system that will solve both the convenience point, and the visual aspect. They are not the first to mention this and I as such I have turned that issue into a general ticket for implementing "bookmarks". These quickmarks are great, but they seem to overlook the other points of this ticket. So it's not that I don't want to merge this, or that a separate ticket is needed to implement "just the quickmarks", but it's more that I want to implement bookmarks properly while doing so.

Either we implement this right, with both quickmark shortcuts and a proper bookmark page with tagging/directories and relevant commands, or we don't. The important part is that we don't have to implement these two separately, which will effectively result in 3 different bookmark implementations. I hope that Vieb can do with just 2: the existing favoritepages + a proper bookmark implementation (potentially this PR), but that will have to include more than just the quickmarks to be sufficient. Hopefully that clears up the confusion.

InconsolableCellist:
It's my fault for not being clear about what I was trying to address about the linked issue, which made it confusing to interpret what I was trying to accomplish with this PR. It should really be relegated to just the quickmarks implementation, which improves the overall bookmark situation but doesn't specifically address all those things in #100.

So with what you said, what's the best way forward? Make a bigger PR over time that implements multiple changes to bookmarks, including but not limited to quickmarks, or to make this PR (or a new one) only address quickmarks, and work on that first? I'd say this feature is almost done, but I haven't considered the other aspects of #100, so I'm not sure how much work a proper solution for those would be.

Jelmerro:
My request is to have a single PR that addresses both quickmarks and #100 in one go, otherwise I think that these features will not feel coherent. And my suggestion for implementation is to start with a nice bookmark page and expand from there. The quickmarks are not the first priority of this ticket, but can definitely be a part of it.

InconsolableCellist:
That makes sense. We should probably close this PR and make it part of a bigger bookmarks feature, though that's a bigger feature to implement than I plan on doing just at the moment

Jelmerro:
Alright, just open a new PR if you ever feel like implementing this or any other issue. Thanks for the effort and the insightful discussion :+1:

PR #115, opened 2020-12-30
Add MacOS big sur themed app icon

SemPruijs:


Jelmerro:
While I like the icon and the idea of a mac themed icon, this PR doesn't seem to do anything with the icon. I would recommend to expand this PR to change the icon automatically for mac users. Erwic mode already allows users to supply a custom icon, and as such can be used as a reference when implementing this. The only difference being that this would need to trigger based on operating system instead of a config file.

Jelmerro:
I'm sorry for automatically closing this due to force pushing, I'm not a fan of that, and hope that this is the first and last time I have to do it. As a result, your PR got automatically closed. Please feel free to make a new one based on the latest master if you ever feel like working on this again.

PR #122, opened 2021-02-07
Fix click-to-insert if site uses stopPropagation()

gebner:
For example, clicking on the search box in duckduckgo did not enable insert mode.

### DIFF
--- a/app/js/preloads/follow.js
+++ b/app/js/preloads/follow.js
@@ -424,7 +424,8 @@ const clickListener = (e, frame = null) => {
         })
     }
 }
-window.addEventListener("click", clickListener)
+window.addEventListener("click", clickListener,
+    {"capture": true, "passive": true})
 
 const contextListener = (e, frame = null) => {
     if (e.isTrusted) {


Jelmerro:
Can we apply the same fix to the other listeners? Or is that not needed? Namely [this one](https://github.com/Jelmerro/Vieb/blob/master/app/js/preloads/follow.js#L448) and [these](https://github.com/Jelmerro/Vieb/blob/master/app/js/preloads/actions.js#L206). I am not familiar with the specifics of stopPropagation.

gebner:
I don't know either, but it's something to remember if we encounter similar issues in the future.  This fix is probably still not bulletproof either, a website could also add a handler with `capture: true`.

I think that `capture: true` should not be added to the context-menu listener, because then I believe it will it also capture click event from `<button>`, etc.

For F11, that's just to prevent the default electron shortcut, right?  The only thing that could happen here is that a website captures F11 itself, not sure if we want to prevent that as well.

Issue #69, opened 2020-09-23
Add command or action to suspend tabs similar to "Tab Suspender"


mfcarneiro:
Hey! First of all, thank you for making this project, I really  enjoy the experience.

The one extension that I always install at first is the tab suspender. I'm wondering if it is possible to add the chrome extensions like Opera does, or will be needed a new feature to do this internally?

Cheers!

Jelmerro:
Custom extensions are not supported, I have looked into this previously, but it still seems like a lot of work to implement this in an easy to use way. I'll have made a [new ticket](https://github.com/Jelmerro/Vieb/issues/70) for extension support in general, for which there surprisingly isn't a ticket yet. I'm hoping that Electron will make this easier in the future, which is a thing they were planning to do. For this ticket I'm interested in the usage of the tab suspend extension. Is that to limit memory and cpu usage by hibernating tabs? I'm not too sure if this is currently possible with Electron, but I'll check if I can find something in the docs.

mfcarneiro:
Hello! 

"Is that to limit memory and cpu usage by hibernating tabs?" yes, pretty much it ;)

I'm not a Electron dev but I found this [electron-chrome-ext](https://github.com/getstation/electron-chrome-extension) Could be helpful?

Thanks!

Jelmerro:
Alright, seems like a nice feature to have regardless of extensions support, I'll check and see what's possible there.

The link you shared seems to be lasted updated in February, so I don't think that it will work in newer Electron versions. Let's move further discussions regarding the extensions to #70.

Jelmerro:
It would seem that all the background suspend features are already enabled by default for Electron's webContents, such as offscreen rendering being disabled and background throttling being enabled.

A more aggressive version I could think of would be:
Closing background tabs after some time, but keep showing it in the tab bar as an "opened" tab, and silently reopen it automatically when you switch back to such a "closed" tab. Keep in mind that this will wipe any forms and other additional changes to the page after it was loaded.

Let me know if such an implementation would be sufficient, and what a sensible timeout duration would be to "suspend" tabs this way.

mfcarneiro:
Thank you very much for focus on this so quickly :smiley:

I will do some tests when using.

Thanks again!

Cheers

Jelmerro:
[...]

PR #123, opened 2021-02-07
Lazy tab restore.

gebner:
At the moment I have 51 open tabs.  Restarting vieb and restoring all the tabs takes a while.

This PR adds support to restore tabs lazily.  That is, tabs that are not pinned or visible are not loaded immediately.  Instead, we set a `lazy` attribute on them.  When switching to a tab with the `lazy` attribute set, the webview is automatically loaded.

(This can also be used to "suspend" tabs as discussed in #69: if you run the `:restart` command with the `restoretabslazily` option enabled, then all tabs except the current one and the pinned tabs will be suspended.)

### DIFF
--- a/app/js/tabs.js
+++ b/app/js/tabs.js
@@ -58,13 +58,17 @@ const init = () => {
                 }
                 if (SETTINGS.get("restoretabs")) {
                     if (Array.isArray(parsed.tabs)) {
+                        const lazy = SETTINGS.get("restoretabslazily")
                         parsed.tabs.map(t => {
                             // OLD remove fallback checks in 4.x.x
                             if (typeof t === "string") {
                                 return {"url": t}
                             }
                             return t
-                        }).forEach(tab => addTab(tab))
+                        }).forEach(tab => addTab({
+                            ...tab,
+                            "lazy": lazy && !tab.pinned
+                        }))
                     }
                     if (Array.isArray(parsed.closed)) {
                         if (SETTINGS.get("keeprecentlyclosed")) {
@@ -178,7 +182,9 @@ const saveTabs = () => {
         data.id = listTabs().indexOf(currentTab())
     }
     listTabs().forEach((tab, index) => {
-        const url = UTIL.urlToString(tabOrPageMatching(tab).src)
+        const webview = tabOrPageMatching(tab)
+        const url = webview.getAttribute("lazy")
+            || UTIL.urlToString(webview.src)
         if (!url || url.startsWith("devtools://")) {
             if (index <= data.id) {
                 data.id -= 1
@@ -214,7 +220,8 @@ const currentTab = () => document.getElementById("current-tab")
 const currentPage = () => document.getElementById("current-page")
 
 const addTab = options => {
-    // Options: url, inverted, switchTo, pinned, container, muted and callback
+    // Options: url, inverted, switchTo, pinned, container,
+    // lazy, muted and callback
     if (!options) {
         options = {}
     }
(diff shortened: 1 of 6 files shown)


gebner:
I just noticed that this causes hangs when searching.

Jelmerro:
There are a lot electron methods that throw an error when the webview dom isn't ready. Most of these cases are patched and checked for regular Vieb use, but I think this change will require a lot more methods to make these safety checks. Though even as is there are scenarios where it will call methods that are unavailable, so maybe this is a nice opportunity to find and catch them as well.

gebner:
Unfortunately it doesn't look like an exception.  Electron just hangs with 100% cpu usage.. Maybe this is related: https://github.com/electron/electron/issues/20602

Jelmerro:
That looks related indeed. I don't think it has much to do with this new feature, but it just started happening because of this new loading mechanism.

It would be cool if we could also use this lazy feature for the actual tab suspend feature, as mentioned in #69, but don't feel pressured to include that in the PR, just mentioning the possibility of implementing it once we have this lazy loading in place.

gebner:
I've now pushed a new implementation which does not use half-initialized webviews.  Instead it first inserts a `<div class="webview">` instead of the actual webview (the class is there so that the CSS can apply to both the suspended and unsuspended versions).  When switching to the tab, the `<div>` is replaced by a `<webview>`.

Suspending is not implemented but should be fairly easy, just replace the `<webview>` by a `<div>` again.

Jelmerro:
That implementation sounds great, I'll look into this PR in a bit more detail later.

app/js/tabs.js
```diff
@@ -317,33 +325,68 @@ const addTab = options => {
...
+}
+
+const unsuspendPage = page => {
+    if (page.tagName === "WEBVIEW") {
+        return
+    }
+    tabOrPageMatching(page).removeAttribute("suspended")
+    const webview = document.createElement("webview")
+    ;["link-id", "container", "class", "id", "style"].forEach(
+        attr => webview.setAttribute(attr, page.getAttribute(attr)))
```
Jelmerro, line 368: I think we can just add muted to this list to read it, and expand this to filter `null` attributes before setting them, like this:
```
;["link-id", "container", "class", "id", "style", "muted"].forEach(attr => {
    if (page.getAttribute(attr)) {
        webview.setAttribute(attr, page.getAttribute(attr))
    }
})
```

app/js/tabs.js
```diff
@@ -317,33 +325,68 @@ const addTab = options => {
...
-    linkId += 1
+    const muted = page.getAttribute("muted") === "true"
+    const url = page.src
+    const callback = page.callback
     webview.addEventListener("dom-ready", () => {
         if (!webview.getAttribute("dom-ready")) {
-            if (options.muted) {
+            const tab = tabOrPageMatching(webview)
+            const title = tab.querySelector("span")
+            if (muted) {
```
Jelmerro, line 389: This if statement can then be changed to check for the attribute, and doesn't need to set it on the tab anymore.

Jelmerro (commented):
Glad we are making progress with this PR, but I have some questions and suggestions.

Jelmerro:
I have squashed this with additional changes into the [V4 branch](https://github.com/Jelmerro/Vieb/compare/feature/v4). Feel free to check there and (if needed) open a new PR to change any issues you might encounter (if any).

gebner:
Sorry, it was a pretty busy past week and I didn't get to the PR.  The changes you made look good, I'm running them now.

Issue #128, opened 2021-02-26
Support for spellcheck of multiple languages simultaneously


TaPO4eg3D:
**Checklist**
- [x] I searched the `:help` documentation using `/` for helpful information
- [x] I have read the [frequently asked questions](https://github.com/Jelmerro/Vieb/blob/master/FAQ.md)
- [x] I did not find any similar issues in the [issue tracker](https://github.com/Jelmerro/Vieb/issues)

**Addition or change**
Currently, `set spelllang` works differently from its Vim counterpart. Vim supports passing multiple languages as well as Electron.
I expect it to work like that: `set spelllang=ru,en_US`

Jelmerro:
Totally agree, for some reason I missed this when implementing this setting. I see that you have already opened a PR for it, I'll leave my feedback on the code there. This feature should definitely be added to Vieb.

Jelmerro:
This is implemented in https://github.com/Jelmerro/Vieb/commit/7b0e8adad86cc939680752d543389c717d1cd224 and will be released this week as part of the V4 release. Great to increase compatibility with Vim by improving the `spelllang` setting, so thanks for this suggestion.

PR #129, opened 2021-02-26
#128 add support of passing multiple options to settings

TaPO4eg3D:
Specifically, add support of multiple languages for `set spelllang`

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -581,11 +581,19 @@ ipcMain.on("set-permissions", (_, permissionObject) => {
 })
 ipcMain.on("set-spelllang", (_, langOptions) => {
     langOptions = new Set(
-        langOptions.split(",").map(lang => lang === "system" ? app.getLocale() : lang.trim())
+        langOptions.split(",").map(lang => {
+            if (lang === "system") {
+                return app.getLocale()
+            }
+
+            return lang
+        })
     )
 
     sessionList.forEach(ses => {
-        session.fromPartition(ses).setSpellCheckerLanguages(Array.from(langOptions))
+        session.fromPartition(ses).setSpellCheckerLanguages(
+            Array.from(langOptions)
+        )
     })
 })
 ipcMain.on("create-session", (_, name, adblock, cache) => {
--- a/app/index.js
+++ b/app/index.js
@@ -579,13 +579,13 @@ ipcMain.on("download-list-request", (e, action, downloadId) => {
 ipcMain.on("set-permissions", (_, permissionObject) => {
     permissions = permissionObject
 })
-ipcMain.on("set-spelllang", (_, lang) => {
+ipcMain.on("set-spelllang", (_, langOptions) => {
+    langOptions = new Set(
+        langOptions.split(",").map(lang => lang === "system" ? app.getLocale() : lang.trim())
+    )
+
     sessionList.forEach(ses => {
-        if (lang === "system") {
-            session.fromPartition(ses).setSpellCheckerLanguages([])
-        } else {
-            session.fromPartition(ses).setSpellCheckerLanguages([lang])
-        }
+        session.fromPartition(ses).setSpellCheckerLanguages(Array.from(langOptions))
     })
 })
 ipcMain.on("create-session", (_, name, adblock, cache) => {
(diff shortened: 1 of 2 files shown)


app/index.js
```diff
@@ -579,13 +579,21 @@ ipcMain.on("download-list-request", (e, action, downloadId) => {
 ipcMain.on("set-permissions", (_, permissionObject) => {
     permissions = permissionObject
 })
-ipcMain.on("set-spelllang", (_, lang) => {
+ipcMain.on("set-spelllang", (_, langOptions) => {
+    langOptions = new Set(
+        langOptions.split(",").map(lang => {
+            if (lang === "system") {
```
Jelmerro, line 585: I like this change, seems much more accurate than setting an empty list (which worked on some systems, but also resulting in no language being checked on others).

app/index.js
```diff
@@ -579,13 +579,21 @@ ipcMain.on("download-list-request", (e, action, downloadId) => {
...
+    )
+
     sessionList.forEach(ses => {
-        if (lang === "system") {
-            session.fromPartition(ses).setSpellCheckerLanguages([])
-        } else {
-            session.fromPartition(ses).setSpellCheckerLanguages([lang])
-        }
+        session.fromPartition(ses).setSpellCheckerLanguages(
+            Array.from(langOptions)
```
Jelmerro, line 595: Why do we need Array.from here? Can't we just build an array to begin with? I don't see the function of building a new set, if we are only going to use it as an array.

app/js/settings.js
```diff
@@ -120,6 +120,9 @@ const listLike = [
     "searchwords",
     "startuppages"
 ]
+const canAcceptMultipleOptions = new Set([
```
Jelmerro, line 123: I would propose the undo the changes to this file, and reimplement them similar to other like-like settings. You can follow the example of `favoritepages` and simply check each element in the list to be a valid language or "system", instead of implementing this new system just for this setting.

app/js/settings.js
```diff
@@ -202,17 +205,28 @@ const init = () => {
...
+    if (canAcceptMultipleOptions.has(setting)) {
+        settingValues = value.split(",")
+    }
+
+    for (const settingValue of settingValues) {
+        const valid = optionList.includes(settingValue.trim())
         if (!valid) {
             const lastOption = optionList.pop()
             const text = `'${optionList.join("', '")}' or '${lastOption}'`
             UTIL.notify(`The value of setting '${setting}' can only be one of:`
```
Jelmerro, line 222: If were going to implement this similar to "favoritepages", it also gives us the option to make this warning more specific and detailed, instead of showing this generic warning.

Jelmerro (commented):
It's great to see that you are implementing your own suggested feature right away, though I do have a few suggestions to make it even better.

Jelmerro:
I have updated the base branch to the new v4 one. You might also want to add your change to the CHANGELOG in the PR, though I can also do that if you don't want to.

app/index.js
TaPO4eg3D: I started with a Set because I wanted to get rid of duplicates like `set spelllocale=en_US, system`. If the system already has the default language as "en_US" it would be duplicated. I don't know how Electron would process it but I rather not risk it. The same applies to passing the Set to Electron function

app/js/settings.js
TaPO4eg3D: Thanks for pointing out the place where I should start to look. It indeed seemed for me like an unnecessary big change for this kind of feature

app/index.js
Jelmerro: Makes sense, but I would move that to happen when setting the option, so that duplicates aren't stored at all, instead of just not being used here.

Jelmerro:
I was experimenting with removing the requirement of opening new tabs when changing the "spell" setting, and while doing that I also implemented support for passing multiple languages to "spelllang". As such, I'm closing this PR and want to apologize for implementing this myself instead of giving you the time to do it. I do want to thank you for suggesting this feature and taking the time to implement it in a PR, it has helped a lot with deciding how to approach this correctly, so thanks!

PR #147, opened 2021-04-10
Clarify pronounciation in readme

brandsjek:


Jelmerro:
Seems like a nice addition, but I would prefer this to be on a separate smaller line, or a bit further down in the readme. What are your thoughts on this?

Jelmerro:
Thanks for the suggestion, but I'll close this PR. I have added the proper pronunciation to the readme in a slightly different way.

PR #148, opened 2021-04-10
Fix splitbelow and splitright flags behavior

TeamTamoad:
Splitright and splitbelow flags seem to be misused. When I set splitbelow flag, it changes the behavior of Vexplore(vsplit) instead of Sexplore(split). When I set splitright flag, it changes the behavior of Sexplore(split) instead of Vexplore(vsplit)


Jelmerro (approved):

Jelmerro:
I have no clue how this was present for so many releases. Though to be fair I never use this myself, but thanks for spotting and fixing this!

TeamTamoad:
You're welcome! This is my first contribution to the open source project, I'm so excited. :smile:

PR #163, opened 2021-04-24
add vimium example viebrc to help

ottopiramuthu:
Please tell me if I am missing anything.

### DIFF
--- a/app/examples/vimium
+++ b/app/examples/vimium
@@ -1,39 +1,30 @@
-map <Esc> <action.toNormalMode>
-map <C-[> <action.toNormalMode>
+" Options
+set adblocker=update
+set downloadmethod=confirm
+set mintabwidth=250
+set norestoretabs
+set redirecttohttp
+set suspendonrestore=regular
+set tabreopenposition=previous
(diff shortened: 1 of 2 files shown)


Jelmerro (commented):
I think it would be better for people using this to only list mappings and settings that are different compared to default. Also, are there no settings that could be updated to match vimium?

app/examples/vimium
```diff
@@ -1,39 +1,30 @@
...
+set downloadmethod=confirm
+set mintabwidth=250
+set norestoretabs
+set redirecttohttp
+set suspendonrestore=regular
+set tabreopenposition=previous
+set windowtitle=title
+set search=https://google.com/search?q=
+set searchwords=w~https://www.wikipedia.org/w/index.php?title=Special:Search&search=%s
+set ignorecase
```
Jelmerro, line 12: Unlike Vim, `ignorecase` and `incsearch` are actually true by default in Vieb, so we can also leave those out. Otherwise this is looking great!

Jelmerro (approved):

PR #170, opened 2021-04-27
Adding a theme to match with macOS Big Sur Dark Mode

bromanko:
I've been using this for a few days and it seems reasonably solid at this point. Figured I'd share in case others were interested.

Nav Bar
image

Mode Selector
image

### DIFF
--- a/app/colors/bigsur-dark.css
+++ b/app/colors/bigsur-dark.css
@@ -1,0 +1,297 @@
+:root {
+  /* general */
+  --bg: #31322a;
+  --fg: #fff;
+  --shadow-color: #444;
+  --overlay-bg: rgba(49, 50, 42, 0.8);
+  --overlay-inner-border: #626262;
+  --overlay-outer-border: #040405;
+  --overlay-backdrop-filter: blur(10px);
+  --text-size: 0.9em;
+  --border-radius: 0.5em;
+  --tab-background: #191914;
+  --tab-scrollbar: #7accd7;
+  --visible-tab: #31322a;
+  --mode-normal-fg: #2d2a2e;
+  --mode-normal-bg: #7accd7;
+  --mode-command-fg: #2d2a2e;
+  --mode-command-bg: #e5c463;
+  --mode-insert-fg: #2d2a2e;
+  --mode-insert-bg: #9ecd6f;
+  --mode-follow-fg: #2d2a2e;
+  --mode-follow-bg: #e5c463;
+  --mode-explore-fg: #2d2a2e;
+  --mode-explore-bg: #ef9062;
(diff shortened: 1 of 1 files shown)


Jelmerro:
I appreciate the effort and I like the looks of the theme. It's nice to have many themes available for users to try without needing to install them manually, but it would be great to have these themes be of interest to a majority of the userbase. For example. the light and flipped theme have been included because a I think a fair share of users would be interested in such a feature, so it seems nice to make this a built-in option. For this theme, it's a little less clear that this would be of interest to many users, so I have mixed feelings about including this. That feeling is amplified by things like: using the same color for two different modes, and also that features such as the navbar color when in explore mode are removed. This is not what I would expect from a built-in theme, so that is what makes me think twice about including user-created themes as built-in ones.

If you're willing to change the theme's colors to better match Vieb's functionality, then I would definitely be able to reconsider. Though for now I would just recommend to share the theme [on reddit](https://reddit.com/r/vieb), either via a new post or the [theme thread](https://reddit.com/r/vieb/comments/kvvgc1/themes_thread/). I hope you understand my consideration for including default themes, and I would also like to hear if you thought of sharing via reddit or other social media before thinking of a PR. I don't want to discourage PRs and I think it's great you went through the trouble of making one, but I also have to make sure that all documented features will continue to work with any built-in themes, and that just isn't the case currently for this colorscheme.

bromanko:
First off, this was no trouble to PR so don't feel any obligation to merge it. I figured I'd share and see if there was interest. I'm also interested in exactly the feedback you offered. What does it take to make a "high-quality" theme? 

* The color catch was a really good one. I didn't even realize I did that!
* What do you mean by the navbar color? The url input and the mode selector change to match the mode. 

Regarding posting to Reddit, I might do that instead. I like the discoverability of themes on Github, that's why I started here. FWIW, there could also be a secondary repo for other themes in the future if you don't want to include too many by default.

Jelmerro:
In explore mode, the border color will change depending on what will happen when you execute `useEnteredData` (by default on `<Enter>`). You can read `:h search` for a detailed explanation. Basically, when you enter only words it will search, so the bar has an orange border. Urls are light blue, files are yellow etc.

A separate repo might be a nice idea, I'll think about that for a while and let you know. I might also consider a dedicated website for sharing colorschemes/viebrc files, but maybe that's a bit overkill for now.

Jelmerro:
What about using [github discussions for it](https://github.com/Jelmerro/Vieb/discussions)? Seems like a nice alternative to reddit that's a bit closer to the repo and the code.

bromanko:
I like that idea. I’ll repost this there after I address your feedback. Thanks!

PR #176, opened 2021-05-02
Quoting the tempfile path for vimcommand

bromanko:
On macOS, the tempfile path contains a space. Lack of quoting caused
the path to be split as a separate arg send to the editor and
prevented opening the file.


Jelmerro:
Thanks for fixing!

Issue #192, opened 2021-05-26
Run custom JavaScript snippets to create or redefine core Vieb functionality kind of like Vim plugins


errge:
**Checklist**
- [x] I searched the `:help` documentation using `/` for helpful information
- [x] I have read the [frequently asked questions](https://github.com/Jelmerro/Vieb/blob/master/FAQ.md)
- [x] I did not find any similar issues in the [issue tracker](https://github.com/Jelmerro/Vieb/issues)

**Addition or change**
I was wondering if there are any plans on making this browser scriptable by somehow loading additional javascript content from `viebrc` (or another config file) and merging them with `actions.js`.  During my initial evaluation of Vieb, I created the following addition in `actions.js`:

```
const openHoplax = () => {
    const {navigateTo} = require("./tabs")
    navigateTo(stringToUrl("file:///home/errge/hoplax/hoplax/index.html?l=/home/errge/docs/hoplax/bookmarks.js"))
    setTimeout(() => currentPage()?.send("focus-input"), 200)
}
```

(We open my custom bookmarks start page, while also focusing the input box on the page automatically.)

(If you want to play around with this website, here is an online copy: https://hoplax.github.io/)

I can maybe find the time to work on this, but before investing any time, I wanted to know your feelings/opinion about slowly turning your project into something that's much more closer to vimscript/emacs than your current simple, small, clean and beautiful design regarding user configuration.

**Alternatives considered**
The alternative I currently use is to just run my own Vieb with `npm start`, but this feels like inferior, e.g. sharing my scripts with my friends would not be so easy.

**Extreme thanks section**
I have to say, this is a *VERY* good browser already.  Still have to use it for a couple of days to have a final opinion, but I'm impressed very much.  I've tried out Nyxt and Qutebrowser too, and they are very interesting and promising, but Vieb is just better.  Being able to git pull, edit the source code a bit and npm start without learning anything more is just amazing.  I have my own patched firefox as my current browser, and patching Vieb was a million times easier.  Thank you very much!

Jelmerro:
#### Language choice
I understand the request of wanting more and more customization, but it has to align with the project to make sense. While I do like JavaScript myself, I can imagine that a lot of die-hard Vim users don't. For the long term interest of Vieb, I therefor think that Vimscript is the only suitable candidate for a scripting language in Vieb.

#### Mappings
[...]

PR #196, opened 2021-05-26
Add open command

errge:
This is very similar to the buffer command, but never tries to be
clever in way of finding a buffer that already shows the webpage.

This is related to #192


Jelmerro:
Thanks for this, it seems like this is much easier than I expected actually.

I didn't put a lot of effort into coming up with the name, and it turns out that `:open` is actually something different in Vim, though they also say it's not really useful, so we might as well use it for this. There is the `:edit` command, though it's not really editing in this case, so I guess I prefer `:open` to `:edit`, even though it's less in line with Vim. What do you think?

One small but important thing you missed is the documentation, please add it to `app/pages/help.html`. This PR is otherwise ready to merge I think.

errge:
I think, I'm fine with open.  To be honest, I think soon you will have people, who are not vim fans, and are still using your project.  Don't tell anyone, but e.g. I use emacs, but are still very impressed with Vieb so far.  So I would consider the importance of the name also in the context of the web, where opening a URL seems like common Jargon.

I added the documentation.

Jelmerro:
I am aware that Vieb has more users than Vim fans by now, but I'm glad to have that confirmed by you as well. I am all ears to make Vieb more user friendly if it doesn't clash with the Vim inspired nature of the project, so I'm happy to have you here as a new contributor :smiley:.

Thanks for the documentation, besides some small references that I will fix shortly this is perfect.

PR #197, opened 2021-05-26
Add two small fixes to verticaltabs.css

errge:
If the tabs are vertical, the pinned tabs don't need to be small,
because the whole row is reserved for them anyways, so we unset the
max-width and width CSS properties.

But this results in a new UI fail: the pinned tabs are not easy to
tell say apart from non-pinned tabs, so to battle this, we add a
slightly darker color as their background (but only if they are not
actually currently selected, in which case we keep the color of the
"cursor").

### DIFF
--- a/app/colors/verticaltabs.css
+++ b/app/colors/verticaltabs.css
@@ -18,6 +18,8 @@
 
 #tabs {overflow-x: hidden;overflow-y: auto;position: absolute;left: 0;width: 15vw;height: calc(100vh - 2em);top: 2em;flex-direction: column;}
 #tabs > span {min-height: 1.9em;max-height: 1.9em;min-width: 0 !important;}
+#tabs > span.pinned {max-width: unset;width: unset;}
+#tabs :not(.visible-tab).pinned {background: #222;}
 #page-container {height: calc(100vh - 2em);position: absolute;top: 2em;left: 15vw;width: 85vw;}
 #app.fullscreen #page-container {height: 100vh;width: 100vw;top: 0;left: 0;}
 #app.tabshidden #page-container {width: 100vw;left: 0;}


Jelmerro:
I do agree that the vertical tabs theme could use improvements, and also specifically this problem is one of the reasons I'm not actually using it myself. Though I'm unsure of a proper solution for it, using the tab color for it seems odd. The tab already gets a darker background when it's suspended in the default theme. I was personally thinking to keep the pinned tabs small, but allow them to go on the same line in the vertical theme, so even there they will take up less space than usual. What do you think?

errge:
I agree with the statement, that this current pull request should not be considered a long-term solution for this problem, because it's not good enough, not beautiful enough and the difference of meanings (pinned, active, suspended) are not showing enough.  I was just annoyed by the small tabs in vertical mode, because they looked like a bug, and I missed seeing the titles. :)

Many icons in one raw would be a usability nightmare for me, because my brain is used to the fact, that when I press a "down" command once on the keyboard, then it goes down one row.  Imagine a vim programming language mode, where j would be rebound to a command that sometimes depending on context goes down, and sometimes goes right.  I think I would not like it.

tab-design

What would you think about this design, that I have in the anim gif? (This is firefox vertical tabs plugin that I currently use.)

Certainly will require a bit more work on my side, as I'm not a CSS expert, but it doesn't look impossibly hard, and looks nice in my opinion. How do you like it?

Maybe I won't have time to do this before the weekend, but I will come back to this and try to do something.  If you prefer to merge this right now to make pinned tabs not absurdly small, then just go ahead, and then I will open another PR with a better UI design.  If you prefer to wait, because it's not urgent for anybody else, then just wait and I will work in this PR.

errge:
Ok, I just realized that the left border is already used in Vieb for indicating things like the muted state of the tab.  What about adding some similar indicator, but on the right border (and do nothing with the background color).  I think all I want is just some visual indication (that looks not ugly, but okay or even beautiful) to show me where are the pinned tabs.

Hmm, another idea: since pinned tabs are always at the top automatically, it might be an idea to add a bottom border below the last one as a kinda separator between pinned and not-pinned?

ottopiramuthu:
Distinguishing between pinned and not-pinned tabs with a separator would be clearer than the other suggestions. Even just a bit more space between the two tab types would be enough. For example, here the first tab is selected and the first two tabs are pinned:
pin

This extra space would only appear when you pin a tab, so I think it would be pretty clear what is happening.

Jelmerro:
I would like to add to this that one of my reasons for pinning tabs was always the reduced amount of space they would take in the tabbar. After investigating, I know at least two other people who pin them for the same reason, so I by default that should also affect the vertical tabs theme, similar to the default one. This will also reduce the amount of things that are suddenly different between built-in themes. Besides, the point of the default themes is not that it's a perfect fit for everyone, but that most people will be able to effectively use and understand it, even if it's a minimal interface. Finally, this is also in line with how Firefox handles pinned tabs, so people migrating to Vieb will be more likely to understand which tabs are pinned and which aren't.

So thanks for the PR, it did in fact make me rethink the current vertical tabs theme and I managed to make it work as described in my initial comment. I'll be closing this PR for the reasons explained above, please try the latest version and let me know if you understand why I would prefer that over both of the suggestions. Besides, the real power of the themes is that you can make your own to make it work for you. As such, here are two example themes to replicate what both of you suggested:

@errge https://gist.github.com/Jelmerro/b2407d3f9afc4e3fc9199c112cb0409a

@ottopiramuthu https://gist.github.com/Jelmerro/8592e531c75128c6b9f09fb3148c3caf

ottopiramuthu:
Both the fix you pushed and the one you made to replicate my suggestion work beautifully on my machine. Your argument for saving space makes sense to me. One reason I prefer keeping titles long is that I disable favicons in my setup. With your wrapping collapsed tabs, I get a bunch of first letters of titles and `...` everywhere:

image

 If you use a lot of pinned tabs I can imagine it being hard to find a lost tab... although that's just a minor nitpick since `:buffer` is more than enough for finding tabs :)

Jelmerro:
That also makes sense, I would recommend using the `:buffer` command if you disable favicons indeed. Glad that the themes now work as expected.

PR #206, opened 2021-06-02
add qutebrowser example viebrc to help

ottopiramuthu:


app/examples/qutebrowser
```diff
@@ -0,0 +1,111 @@
...
+nmap Ss <:help>
+nmap <C-A-p> <:print>
+nmap <C-q> <:quit>
+nmap <C-W> <:quit>
+nmap ZQ <:quit>
+nmap gd <:write>
+nmap <C-Tab> <action.nextTab>
+nmap <C-S-PageUp> <action.moveTabBackward>
+nmap <C-S-PageDown> <action.moveTabForward>
+nmap <C-S-Tab> <action.previousTab>
```
Jelmerro, line 45: First you set it here, then you set it to `nop` below, can't we just leave both of these lines out? This mappings isn't mapped by default anyway.

app/examples/qutebrowser
```diff
@@ -0,0 +1,111 @@
...
+cmap <C-?> <Del>
+cmap <C-h> <BS>
+cmap <C-b> <Left>
+cmap <C-f> <Right>
+cmap <C-k> <S-End><BS>
+cmap <C-u> <S-Home><BS>
+cmap <C-w> <C-S-Left><BS>
+
+pmap c <action.toNormalMode>
+pmap e <pointer.moveFastRight>
```
Jelmerro, line 102: I understand mapping this to e due to this being similar to Vim, so this is fine, but we should still provide the option to inspect at a position. Is there something similar in qutebrowser that we can map to a different key?

app/examples/qutebrowser
```diff
@@ -0,0 +1,111 @@
...
+set fontsize=16
+set norestoretabs
+set permissioncamera=ask
+set permissiondisplaycapture=ask
+set permissionmediadevices=allow
+set permissionmicrophone=ask
+set permissionpersistentstorage=allow
+set redirecttohttp
+set tabclosefocusright
+set tabreopenposition=next
```
Jelmerro, line 15: This can only be set to 'left', 'right' and 'previous'.

app/examples/qutebrowser
```diff
@@ -0,0 +1,111 @@
...
+nmap <A-7> <:b 6>
+nmap <A-8> <:b 7>
+nmap <A-9> <:b 9999>
+nmap <C-^> <action.toLastUsedTab>
+nmap <C-Tab> <action.toLastUsedTab>
+nmap ga <action.openNewTab>
+nmap gC <action.openNewTabWithCurrentUrl>
+nmap gJ <action.moveTabForward>
+nmap gK <action.moveTabBackward>
+nmap <C-w> <:close>
```
Jelmerro, line 66: This will prevent all window split mappings from being accessed, as those are prefixed with `<C-w>`, we might want to reconsider including this even if it's possible in qutebrowser.

app/examples/qutebrowser
ottopiramuthu: There is only a command to open devtools ([devtools](https://qutebrowser.org/doc/help/commands.html#devtools)) and a command to focus devtools ([devtools-focus](https://qutebrowser.org/doc/help/commands.html#devtools-focus)). I don't know of any way to inspect at a position.

Jelmerro: In that case this is fine.

app/examples/qutebrowser
ottopiramuthu: Hmm.. Maybe I could comment it out?

app/examples/qutebrowser
ottopiramuthu: I will set it to 'right'. I confused it with qutebrowser's next/previous.

app/examples/qutebrowser
ottopiramuthu: Good point. Confusingly, qutebrowser sets these both by default.

app/examples/qutebrowser
```diff
@@ -0,0 +1,111 @@
...
+nmap = <action.zoomReset>
+nmap <C-p> <:pin>
+nmap <A-m> <:mute>
+nmap <C-s> <action.stopLoadingPage>
+nmap o <action.toExploreMode><BS>
+nmap go <action.toExploreMode><End>
+nmap gt <action.toCommandMode>buffer<Space>
+nmap O <action.openNewTab><action.toExploreMode>
+nmap yy <action.pageToClipboard>
+nmap pp <action.openFromClipboard>
```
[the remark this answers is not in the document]
ottopiramuthu: That makes sense. Will do.

app/preload/help.js
```diff
@@ -146,7 +146,7 @@ const processHash = () => {
 window.addEventListener("hashchange", processHash)
 
 window.addEventListener("DOMContentLoaded", () => {
-    for (const example of ["chromium", "firefox", "vimium", "tridactyl"]) {
+    for (const example of ["chromium", "firefox", "qutebrowser", "vimium", "tridactyl"]) {
```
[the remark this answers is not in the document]
ottopiramuthu: This looks cleaner. I will change it.

app/examples/qutebrowser
Jelmerro: That would be alright I think, just make sure to also add a comment above as to why it's disabled, you can refer the user to the following section for a detailed explanation:  `:h spitting`

app/examples/qutebrowser
Jelmerro: Actually, now there is a mapping for going to the next tab with `<C-Tab>`, but you can't go back with `<C-S-Tab>`, I was mistaken before, we should just have removed the nop line and leave this one as is I think.

app/examples/qutebrowser
ottopiramuthu: I've added the comments. Feel free to suggest changes to the wording.

app/examples/qutebrowser
ottopiramuthu: I've readded the mapping.

Jelmerro:
Just one more thing I noticed after testing it, is there no way to click at the pointer position in qutebrowser? Because you seem to have overwritten `<CR>`, which in itself is fine, but I suspect there must be another way to click at the pointer position?

ottopiramuthu:
Would moving the `<CR>` mapping to `vmap` work? I figure you would only want to copy text in visual mode.

Jelmerro:
Yeah, that would work, copying text in pointer mode isn't actually that relevant indeed. Just make that a mapping in visual and leave `<CR>` to click while in pointer.

Jelmerro:
Thanks again!

ottopiramuthu:
Happy to help :)

PR #207, opened 2021-06-04
add custom commands to qutebrowser viebrc

ottopiramuthu:


Jelmerro (commented):
I really like this PR, I think this will be much more familiar for qutebrowser users. Though I have a question about `open_-t_--_{clipboard}` (and other commands with curly brackets): can you link me the docs in qutebrowser for this? Because the name is fairly complex and I don't really understand why we need all those special characters. Is it a placeholder for a command argument in qutebrowser perhaps?

ottopiramuthu:
For the clipboard command check out the open command:
 https://qutebrowser.org/doc/help/commands.html#open

Basically all it's doing is opening the contents of the clipboard in a new tab.

The arguments in curly braces can be found at the top of the commands page:

https://qutebrowser.org/doc/help/commands.html

in the first section (commands).

I apologize for the formatting, I'm writing this on a phone.

Jelmerro:
What you are saying makes sense, but I don't really find the ones with the double dashes and curly brackets more readable than the regular Vieb syntax. I'm not sure if we should change those, what do you think about it?

ottopiramuthu:
I think the people who will use this viebrc will be familiar with most of Qutebrowser's commands. They will want Qutebrowser's syntax for as many mappings as possible. Regarding readability, Vieb's commands are all more readable since there are no flags so they tend to be self-descriptive (e.g. `open -t` vs `openNewTab`). If we prioritize readability, no Qutebrowser commands will be left. I prefer keeping those complex-looking commands for this reason (and for consistency with the other custom commands).
If you really want more readable commands, I can make a commit that uses long versions of flags everywhere (e.g. `--tab` instead of `-t`).

Jelmerro:
I definitely understand what you are saying, let's just use the long version of flags in that case :slightly_smiling_face:

Jelmerro:
Thanks a lot for this!

Issue #83, opened 2020-11-03
docs: Compare to plugins like Vimium and Tridactyl


lorenzleutgeb:
I am genuinely interested in Vieb. At the moment I use Firefox with Vimium and TreeStyleTabs, and I have tried Tridactyl. There are other similar plugins available.

I think it would be beneficial for adoption of Vieb if you could provide
 - [ ] a comparison of Vieb versus plugin-based solutions like the one I am using now, and
 - [ ] migration guides.

Jelmerro:
## Motivation

The main reason me to start Vieb was that most plugins at the time that I tried just emulated a few keys and called it a Vim plugin. With Vieb the plan is to make it FEEL a lot more like Vim than the plugins ever did for me. I haven't tried every single plugin, and there might very well be plugins that do this properly. I never did find them, and as such I made Vieb and use it daily. Here are a few things I remember disliking about the plugins back then.

### Separate modes

This meant that shortcuts often conflicted with the website shortcuts. A big part of the initial design of Vieb was the clear separation of modes, similar to Vim.

In Vieb, I haven't had an issue like this, because I only use insert mode to type text and very rarely to do other stuff (again, similar to Vim). The mode of Vieb is prominently displayed in the navbar, so there's never any confusion what the pressed keys will do.

### Follow mode

Because of the issue above, picking a follow mode item is inconsistent, and sometimes has side-effects. Besides that, most plugins weren't very clear about where it would actually click, as they would show the keys on the topleft of the element, usually obstructing the view to what you would actually click on. Finally, it was fairly unclear what would happen upon clicking something, because you have to assume if it's a button, a link or something else.

In Vieb, a box is shown around the element so you know which element it is and the expected keys are on the topright so they don't overlap the element you wanted to click on. The links are all color coded based on type, and you can choose to open links in the current tab or in a new tab.

### Vim-plugin limitations

Due to the nature of browser plugins, many proper integrations that you would expect just aren't possible with web extensions. I don't remember any of them having very useful settings or mappings, but it's also not possible (at least in Firefox) to modify all browser settings from a plugin.

Vieb has `:set` and `:map` commands that I personally can't distinguish from Vim in terms of syntax. I'll just end with that for now, otherwise this comment might get even longer.

### What else

[...]

PR #208, opened 2021-06-06
add custom commands to vimium viebrc

ottopiramuthu:
This adds custom commands from Vimium and some mappings that were forgotten in the initial draft because they were hidden under "advanced commands". `F` also now correctly opens tabs in the foreground. Commands from Vimium that are available in Vieb but aren't mapped in Vimium are also included in a separate paragraph.

### DIFF
--- a/app/examples/vimium
+++ b/app/examples/vimium
@@ -3,7 +3,7 @@ set adblocker=update
 set downloadmethod=confirm
 set mintabwidth=250
 set norestoretabs
-set follownewtabswitch
+set nofollownewtabswitch
 set redirecttohttp
 set tabreopenposition=previous
 set windowtitle=title
--- a/app/examples/vimium
+++ b/app/examples/vimium
@@ -3,6 +3,7 @@ set adblocker=update
 set downloadmethod=confirm
 set mintabwidth=250
 set norestoretabs
+set follownewtabswitch
 set redirecttohttp
 set tabreopenposition=previous
 set windowtitle=title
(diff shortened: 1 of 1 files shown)


app/examples/vimium
```diff
@@ -9,26 +9,93 @@ set windowtitle=title
 set search=https://google.com/search?q=%s
 set searchwords=w~https://www.wikipedia.org/w/index.php?title=Special:Search&search=%s
 
+" Custom Commands
+command! enterInsertMode <toInsertMode>
+command! enterVisualMode <p.start><p.startVisualSelect>
+command! focusInput <insertAtFirstInput>
+command! LinkHints.activateMode <startFollowCurrentTab>
+command! LinkHints.activateModeToOpenInNewTab <startFollowNewTab><toLastUsedTab>
```
Jelmerro, line 17: What exactly is this supposed to do? It doesn't seem to work here. You might be looking for the `follownewtabswitch` setting. If so, please use that in the example instead of this mapping.

Jelmerro (commented):
This is really nice, just one small comment.

app/examples/vimium
ottopiramuthu: Yes this is supposed to switch to the followed tab. I have changed it in the following commit.

app/examples/vimium
```diff
@@ -3,6 +3,7 @@ set adblocker=update
 set downloadmethod=confirm
 set mintabwidth=250
 set norestoretabs
+set follownewtabswitch
```
Jelmerro, line 6: This will enable the toggle, which is already the default, I think you are looking to disable it instead? You can do this using `set nofollownewtabswitch`.

ottopiramuthu:
Do you think it could be a good idea to map keys for functionality that is not in Vieb to `<nop>`, so we can easily track what bindings from the example plugins/browsers are missing in Vieb.

Jelmerro:
I would personally prefer to use guides to outline what is and isn't supported. I don't think it would really benefit the example configs  to be filled with a list of `<nop>` calls.

As for this PR, it seems ready to merge, so I'll merge it and we can discuss a new ways to improve it [here, where more people will read it](https://github.com/Jelmerro/Vieb/issues/83).

PR #226, opened 2021-06-14
add configuration fragment: digraphs

ottopiramuthu:
For the Javascript I pretty much copy-pasted the code from the example buttons. Let me know if anything is wrong or needs to be changed.


Jelmerro:
I'm sorry to disappoint, but it seems that Electron [doesn't support this](https://www.electronjs.org/docs/api/structures/keyboard-input-event) and these mappings also don't work for me as a result. There was an issue in Electron that asked for something similar, but it was [closed](https://github.com/electron/electron/issues/23052) as won't fix sadly. I really appreciate your work on this, but it won't be of use until Electron suddenly decides to implement this.

PR #230, opened 2021-06-16
add pentadactyl viebrc to help

ottopiramuthu:
Let me know if you want to keep `e` for visual/pointer mode. Inspect element is available in Pale Moon, but not through Pentadactyl (you have to right-click, there is no command).


Jelmerro (approved):

Jelmerro:
It's fine like this, thanks!

PR #231, opened 2021-06-16
add vim vixen viebrc to help

ottopiramuthu:


app/examples/vim vixen
```diff
@@ -0,0 +1,55 @@
...
+
+" Mappings
+nmap 0 999<scrollPageLeft>
+nmap $ 999<scrollPageRight>
+nmap o <toExploreMode><BS>
+nmap o <toExploreMode>
+nmap t <openNewTab><toExploreMode>
+nmap T <openNewTabWithCurrentURL>
+nmap b :buffer<Space>
+nmap D <:close><previousTab>
```
Jelmerro, line 40: If i understand correctly, this makes `d` focus the tab on the right, while `D` focuses the one on the left, similar to pentadactyl?

app/examples/vim vixen
```diff
@@ -0,0 +1,55 @@
...
+set searchwords+=bing~https://www.bing.com/search?q=%s
+set searchwords+=duckduckgo~https://duckduckgo.com/?q=%s
+set searchwords+=twitter~https://twitter.com/search?q=%s
+set searchwords+=wikipedia~https://en.wikipedia.org/w/index.php?search=%s
+
+" Mappings
+nmap 0 999<scrollPageLeft>
+nmap $ 999<scrollPageRight>
+nmap o <toExploreMode><BS>
+nmap o <toExploreMode>
```
Jelmerro: Should one of these be `O` instead of `o`?

app/examples/vim vixen
ottopiramuthu: Yes, that is what it does.

ottopiramuthu: Should I change it to something like what I did here: `nmap !d <:set closablepinnedtabs><:close><:set noclosablepinnedtabs>`. That would avoid briefly being on the wrong page.

Jelmerro: How about:
```
nmap d <:set tabclosefocusright><:close>
nmap D <:set notabclosefocusright><:close>
```

Jelmerro: We can also fix this for the pentadactyl example. Finally, would you mind renaming the file to `vimvixen` without the space? I"m not too sure Windows will like that :sweat_smile:

ottopiramuthu: I used a space in the file because I didn't want the button in the help to have "Vimvixen" when the name of the plugin is "Vim vixen". I can rename it, but would you consider changing the parsing to underscores become spaces on the button?

Jelmerro: How about removing spaces when looking up the filename?

ottopiramuthu: Do you mean in `examples [ ... ]` it would be `vim vixen`, but then remove the spaces in 
```
link.href = joinPath(__dirname, `../examples/${example}`)
```

ottopiramuthu: I've tried to do this in the latest commit. I will fix tridactyl and any other viebrcs that have similar tab close mappings in a separate PR.

Jelmerro: That's nice, but maybe also rename the file to match your changes?

ottopiramuthu: I have renamed the file in the latest commit.

Jelmerro (approved):

PR #232, opened 2021-06-16
fix pentadactyl tab close mappings

ottopiramuthu:


Jelmerro:
I only just noticed your PR, I already made a similar one myself. If there is anything else that needs changes, feel free to open a new one based on the recent changes.

PR #233, opened 2021-06-16
update cheatsheet with visual mode text mappings

ottopiramuthu:


Jelmerro:
Really great that you are so fast in updating the cheatsheet, this would have taken me hours (I don't use inkscape often :sweat_smile:).

Should we add `*` to the list as well, maybe on the same line as `ts` and just show both as a mapping for that action?

ottopiramuthu:
Oh believe me, it took me hours when I first started this XD

ottopiramuthu:
I've added `*` in the latest commit.

Jelmerro:
Thanks!

Issue #57, opened 2020-06-01
Per-site pass through of keybindings


ipod825:
Many of google's page has their own mappings. For e.g. in gmail, you can configure it to press `j`, `k` to check next and previous email. In calendar, you can press `d`,`w`,`m`,`a` to switch between day/week/month/agenda mode. As vieb grabs all the keybindings, these site-dependent keybindings no longer works.

Jelmerro:
One of the (many) core reasons for developing Vieb, for me, was the lack of separation between Vim-like keybindings and the website shortcuts. I have never regarded this merging of shortcuts as a positive, and therefor I haven't considered the possibility of adding this to Vieb.

Having said that, I'm not opposed to adding per-site pass-through of specific shortcuts. One of the things that isn't clear to me is how this would be configured in way that is in line with Vieb's current setting/mapping structure.

ipod825:
Perhaps you can add a function called `passthroughkey`
```vim
nmap j call passthroughkey("j", "https://site-that-need-to-pass-through")
```

Implementation of `passthroughkey` would then be something like:
```vim
function! passthroughkey(key, site)
if match(current_url, site)
 " pass through a:key
else
  " handle key in vieb
endif
```

Jelmerro:
I agree that this is a solution that would fit a Vim-like browser, but custom functions are not supported by Vieb as of today. Therefor we might want to consider a different implementation for now or we will need to wait for functions to be supported by Vieb before implementing this pass through feature.

ipod825:
I think in addition to per-site pass-through. There should also be a global pass-through.
Regarding implementation complexity, I think these could be set using list, which is easier to parse.

```vim
let global_passthrough_keys = ['<Left>', '<Right>', '<Space>']
let per_site_passthrough_keys = [['j', _'*mail.google.com/*']]_
```

Jelmerro:
I would like the set command to remain true to Vim, but we could introduce new commands to manage the passthough dictionary. I would probably stick to one dictionary and allow users to make global passthroughs by supplying an all-matching pattern as the domain.

ipod825:
> I would like the set command to remain true to Vim

What does this mean? I thought mean also have options taking list/dict as argument.

> I would probably stick to one dictionary and allow users to make global passthroughs by supplying an all-matching pattern as the domain.

Better!

Jelmerro:
[...]

PR #240, opened 2021-06-17
add saka key viebrc to help

ottopiramuthu:
Saka key has a bunch of mappings that conflict with Vieb mappings to enter modes. Some of them (e.g. `i`) have alternative default mappings, but others (e.g. `vv`) do not. The list of mappings starts [here](https://key.saka.io/docs/tutorial/basics). Let me know what you want to change.

### DIFF
--- a/app/examples/sakakey
+++ b/app/examples/sakakey
@@ -1,0 +1,82 @@
+" Options
+set adblocker=update
+set devtoolsposition=split
+set downloadmethod=confirm
+set firefoxmode=always
+set fontsize=16
+set mintabwidth=250
+set norestoretabs
+set permissioncamera=ask
+set permissiondisplaycapture=ask
+set permissionmediadevices=allow
+set permissionmicrophone=ask
+set permissionpersistentstorage=allow
+set redirecttohttp
+set splitbelow
+set tabclosefocusright
+set tabreopenposition=previous
+set windowtitle=title
+set noshowcmd
+set notimeout
+
+" Mappings
+nmap d <scrollDown>
+nmap s <scrollUp>
+nmap J <scrollPageDown>
+nmap K <scrollPageUp>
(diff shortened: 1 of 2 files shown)


Jelmerro:
Looks pretty good, besides the missing mappings to go into insert mode. The Sara key project seems to have [these pass keys](https://key.saka.io/docs/tutorial/pass-keys). The `;` mapping might make sense to map after #57 is implemented, but `:` is problematic as a shortcut for going to insert mode ("shift-;" is how they list it). There is also a problem with `ctrl-x` for leaving insert mode, as that is `Cut` on most systems, so I'm not sure why these common keys for regular functionality are mapped to such different keys. What do you think makes sense as someone who has used Sara Key more than I have?

ottopiramuthu:
Remapping `:` is unacceptable in my view, so I would make the following changes to Saka key's mappings:
- `:` becomes `<A-;>`
- `ctrl-x` becomes `<A-x>`

Alt is already used in some places, and giving passing-related keys a special modifier so they don't conflict with standard shortcuts seems good to me. I don't use passing keys often in any Vim browser, so this small change wouldn't affect me much.

Let me know what you think.

Jelmerro:
That seems like a decent compromise, let's go with that :+1:

ottopiramuthu:
I'd be happy to add these mappings, but I'm not sure how to make mappings in Vieb that act like `:` and `ctrl-x`.. and I'm not sure from the pass key issue how these would be implemented in Vieb. Should I just use insert mode?

Jelmerro:
Yes, that's what is the closest alternative to passthrough mode in Vieb, just add those two mappings and I think it's alright.

ottopiramuthu:
I've added those mappings in the latest commit.

Jelmerro:
Great, thanks!

PR #241, opened 2021-06-20
add vivaldi viebrc to help

ottopiramuthu:
I've changed their `<Esc>` to stop loading the page to `<S-Esc>`.

### DIFF
--- a/app/examples/vivaldi
+++ b/app/examples/vivaldi
@@ -76,7 +76,7 @@ nmap <S-F3> <previousSearchMatch>
 nmap <BS> <backInHistory>
 nmap <C-Left> <backInHistory>
 nmap <A-Left> <backInHistory>
-nmap <S-Esc> <stopLoadingPage>
+nmap <Esc> <stopLoadingPage>
 nmap <C-s> :write<Space>
 
 " vim: ft=vim
--- a/app/examples/vivaldi
+++ b/app/examples/vivaldi
@@ -10,7 +10,6 @@ set permissionmediadevices=allow
 set permissionmicrophone=ask
 set permissionpersistentstorage=allow
 set redirecttohttp
-set search=https://google.com/search?q=%s
 set splitright
 set suspendonrestore=none
 set tabreopenposition=previous
--- a/app/examples/vivaldi
+++ b/app/examples/vivaldi
@@ -79,4 +79,6 @@ nmap <A-Left> <backInHistory>
 nmap <Esc> <stopLoadingPage>
 nmap <C-s> :write<Space>
 
+cmap <F2> <toNormalMode>
+
 " vim: ft=vim
--- a/app/examples/vivaldi
+++ b/app/examples/vivaldi
@@ -1,0 +1,83 @@
+" Options
+set adblocker=update
+set cache=full
+set devtoolsposition=vsplit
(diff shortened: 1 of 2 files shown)


app/examples/vivaldi
```diff
@@ -0,0 +1,83 @@
...
+nmap <C-e> <toExploreMode><BS>
+nmap <C-f> <toSearchMode>
+nmap <C-g> <nextSearchMatch>
+nmap <F3> <nextSearchMatch>
+nmap <C-G> <previousSearchMatch>
+nmap <S-F3> <previousSearchMatch>
+nmap <BS> <backInHistory>
+nmap <C-Left> <backInHistory>
+nmap <A-Left> <backInHistory>
+nmap <S-Esc> <stopLoadingPage>
```
Jelmerro, line 80: `<Esc>` is actually unmapped by default in normal mode, so this can be just be `<Esc>` I think.

app/examples/vivaldi
```diff
@@ -0,0 +1,83 @@
...
+nmap <Up> <scrollUp>
+nmap <Down> <scrollDown>
+nmap <Left> <scrollLeft>
+nmap <Right> <scrollRight>
+map <PageUp> <scrollPageUp>
+map <PageDown> <scrollPageDown>
+map <Home> <scrollTop>
+map <End> <scrollBottom>
+map <F3> <toSearchMode>
+nmap <F2> <toCommandMode>
```
Jelmerro, line 32: Can we map these keys in normal mode only? `PageDown` etc. have very different functions when typing text for example.

Also, are the `<F2>` and `<F3>` keys toggles for switching back and forth between these modes? I that case it would be nice to also add these mappings to search and command mode to go back to normal mode.

app/examples/vivaldi
```diff
@@ -0,0 +1,83 @@
...
+set permissionpersistentstorage=allow
+set redirecttohttp
+set search=https://google.com/search?q=%s
+set splitright
+set suspendonrestore=none
+set tabreopenposition=previous
+set windowtitle=title
+
+" Search Engines
+set search=https://www.bing.com/search?q=%s
```
Jelmerro, line 20: Right above you set the search to google, which of the two should it be?

app/examples/vivaldi
ottopiramuthu: `<F2>` is a toggle. `<F3>` is not. I've changed this in the latest commits.

app/examples/vivaldi
ottopiramuthu: It is bing by default. This is fixed in the latest commits.

app/examples/vivaldi
ottopiramuthu: I missed this and fixed it in the latest commits.

app/examples/vivaldi
ottopiramuthu: I need to change the behavior of `<F3>`- and `<C-g>`-related commands to better match Vivaldi. I will commit those changes once they work.

Jelmerro: Great, let me know when I can check it again.

ottopiramuthu: I've fixed this in the latest commit.

PR #242, opened 2021-06-20
fix searching mappings for firefox and chromium

ottopiramuthu:
This makes the same fixes to searching made in the Vivaldi viebrc, since searching behaves the same in the other two mainstream browsers.


Jelmerro:
Thanks, I have also updated the examples with the same fix related to PageUp etc. that I suggested for the other example just now.

PR #248, opened 2021-06-22
update cheatsheet for 5.2.0

ottopiramuthu:
Added `gv`, `.`, and `@:`


Jelmerro:
Nice, but there is something that I think we can still improve. There are now two different "misc. actions", one you just added, and the one below the map commands. Could these two be merged at either of the locations, since these are both just a single mapping?

ottopiramuthu:
I've fixed this in the latest commit.

Jelmerro:
Thanks!

PR #250, opened 2021-06-24
fix qutebrowser example flags

ottopiramuthu:
I forgot to use the long version of flags.


Jelmerro:
I forgot to check this, because `-t` was actually pretty clear to me, but the long version is better for consistency, so thanks!

PR #285, opened 2021-08-29
update cheatsheet

ottopiramuthu:


Jelmerro:
Great, thanks again! Though if it's not too much to ask, could you also add the changes from [this commit](https://github.com/Jelmerro/Vieb/commit/1f6cdaf28f4c32930349afb829ce06a44ce6fe16) regarding the newtab actions and pointer mode searching?

ottopiramuthu:
I will add those in a new PR, since they are a different kind of change.

Jelmerro:
There is something wrong in the new PNG on the bottom right, it's missing a few combinations and the `di` example that was there previously.
image
image
Besides that, the intended changes are all good.

ottopiramuthu:
Adding a new line made the "di... " text go beyond the bounding box. This should be fixed in the latest commit.

Jelmerro:
Sorry, next problem :joy: Now only the pointer mode section is in the PNG...

ottopiramuthu:
Fixed in the latest commit.

Jelmerro:
That's much better already, just a tiny thing left, which is that the spacing was reduced around the sheet. I would prefer it back, as it makes integrating it into other places easier and it looks nicer around the edges where the text is now being cut off in the middle of a character.
image

ottopiramuthu:
Not sure how that changed; the latest commit should fix this.

Jelmerro:
It does not sadly, instead of having a bit of spacing around the edges, it seems to be zoomed in just a bit by using the original dimensions. I think the diff below is the cause.
image

ottopiramuthu:
I've added some space in the latest commit.

Jelmerro (approved):
Now it's all good, thanks!

PR #292, opened 2021-09-02
add menu actions to cheatsheet

ottopiramuthu:


Jelmerro:
I know it's not actually related to this change, but it would be great if we could also list `<Enter>` for selecting a menu item there. What do you think?

ottopiramuthu:
Good idea. I've added it in the latest commit.

Jelmerro:
Thanks a lot!

PR #301, opened 2021-09-08
add screenshot to cheatsheet

ottopiramuthu:


Jelmerro:
Nice, but for some reason the border got removed again...

ottopiramuthu:
The latest commit has added space around the border.

Jelmerro:
Are you sure about that, because I don't see it in the PNG. The reason for this seems to be the same as last time, if you check the [diff](https://github.com/Jelmerro/Vieb/pull/301/files) you can see the dpi and sizes are being changed.

ottopiramuthu:
The spacing has changed again in the latest commit. When I open the png in GIMP, for example, there is space around the text.

PR #303, opened 2021-09-09
add screenshot mappings to examples

ottopiramuthu:
Adds screenshot mappings to the example viebrcs. Chrome, Vivaldi, and Qutebrowser all have ways to screenshot the page, but they are not default keyboard shortcuts, so I left them out.

### DIFF
--- a/app/examples/firefox
+++ b/app/examples/firefox
@@ -47,6 +47,7 @@ map <A-6> <:b 5>
 map <A-7> <:b 6>
 map <A-8> <:b 7>
 map <A-9> <:b 8>
+map <C-S> <:screenshot>
 vmap <C-f> <p.searchText>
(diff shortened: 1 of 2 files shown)


Jelmerro:
What way do other browsers have for screenshotting the page? I can imagine that especially qutebrowser would just have a command for it that we could include in the examples. If not, we can merge this PR as is.

ottopiramuthu:
Qutebrowser has the command `:screenshot`. Vivaldi has an action to capture the screen, but there is no binding to it. Chrome has extensions and you can capture using [the devtools](https://www.computerworld.com/article/3408017/capture-screenshots-chrome.html).

Jelmerro:
Makes sense, thanks for adding the missing shortcuts to the examples.

PR #315, opened 2021-09-15
rearrange qutebrowser custom commands

ottopiramuthu:
Reorders the custom commands for qutebrowser for clarity.
Motivated by the suggestions in [this video](https://www.youtube.com/watch?v=nWUk4tMLUBA).

### DIFF
--- a/app/examples/qutebrowser
+++ b/app/examples/qutebrowser
@@ -17,187 +17,113 @@ set windowtitle=full
 set mapsuggestposition=bottomleft
 
 " Custom Commands
-command! scroll_up <scrollUp>
-command! scroll_down <scrollDown>
-command! scroll_left <scrollLeft>
-command! scroll_left_ <p.scrollLeft>
-command! scroll_right_ <p.scrollRight>
-command! scroll_right <scrollRight>
-command! scroll-page_0_-1 <scrollPageUp>
-command! scroll-page_0_1 <scrollPageDown>
-command! scroll-to-perc_0 <scrollTop>
-command! scroll-to-perc <scrollBottom>
-command! tab-prev <previousTab>
-command! tab-next <nextTab>
-command! reload_--force <reloadWithoutCache>
-command! set_ <:help>
-command! download <:write>
-command! undo <reopenTab>
-command! navigate_prev <previousPage>
-command! navigate_prev_--tab <previousPageNewTab>
-command! navigate_next <nextPage>
-command! navigate_next_--tab <nextPageNewTab>
-command! navigate_up <toParentUrl>
-command! navigate_up_--tab <toParentUrl><openNewTabWithCurrentUrl><CR><toLastUsedTab><backInHistory><toLastUsedTab>
-command! tab-focus_1 <:b 0>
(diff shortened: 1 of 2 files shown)


Jelmerro (commented):
Much better like this, but I also hear the feedback in the video that these custom commands don't actually make a lot of sense to have. They suggest to just map the action directly with the map commands, as to reduce the amount of config in the examples. What do you think?

ottopiramuthu:
I added the custom commands so the hints that show up in `mapsuggest` look as similar as possible to what one would see in Qutebrowser. I notice the person in the video isn't using the suggestions, so it makes sense that they would see this intermediate step as unneeded/confusing. I'd be happy to remove the custom commands, but keep in mind the reason they were added in the first place. 

The one other issue I saw from the video was that the arrow keys are mapped to `<commandHistoryNext/Prev>` in command mode. That mirrors the name given to it in Qutebrowser (`completion-item-focus --history next`). From what I can tell, Qutebrowser goes up/down in the command history when the prompt is empty (`:`), otherwise it uses the suggestions (e.g. `:open `).  The person in the video likely rarely uses the arrows when the prompt is empty in Qutebrowser, so they were frustrated they didn't map to `prev/nextSuggestion`. Since there isn't (yet?) a way to check for the contents of the navbar in mappings, I don't know what to do with this.

Jelmerro:
I agree that we shouldn't remove them completely and as such propose to split up the examples into one that does use these command names and one that doesn't: `examples/qutebrowser` & `examples/qutebrowser-customcommands` or something, so that we don't have force these commands on people just looking to get into Vieb that are okay with using a slightly different name for some of the actions to have a slimmer/cleaner viebrc config. Regarding they arrows, I guess it makes more sense to have them scroll the suggestions instead of the history if that only happens when the url box is empty, as people will usually type before using the arrow keys (unless they know what they are doing, in which case they know to use `<C-n>` and `<C-p>` for navigating the command/explore history..

ottopiramuthu:
The most recent commits fix the arrow key mappings and split the config into two files. Let me know if you want me to add `qutebrowser-custom-commands` to the help.

Jelmerro (approved):
This is great, thanks a lot. If you ever find the time, we could eventually do the same for the other configs with custom commands.

ottopiramuthu:
I'll probably do the other configs over the weekend.

nephitejnf:
I know I am a week late looking at this, but is there a way for Vieb to make custom commands with spaces. One of the biggest problems that I saw, when it comes to the custom commands, was that the underscores created a keystroke that is out of the way if you were to type them in. I get it likely isn't a feature right now. It could also allow for mapping arguments to native commands directly. So when you get a numerical, it passes directly to the command native command you are mapping to, instead of creating nine to ten commands for each number.

Jelmerro:
You are right, there is no such thing in Vieb, which is exactly why it's better to by default focus on what Vieb CAN do instead of using commands that kind of look like another project and can't just get it to match exactly. I welcome PRs to improve custom commands to be more like Vim, but I won't pursue a command rework in the near future myself.

PR #318, opened 2021-09-16
add custom-commands configs for examples

ottopiramuthu:
I wrote a python script to do this for me; you don't have to wait for the weekend ;)


Jelmerro (approved):
Thanks, cool that you used a script for it as well.

PR #330, opened 2021-11-11
Enabled searchword to accept multiple queries

### LINKED ISSUE #326: Improve `searchwords` %s
**Checklist**
- [x] I searched the `:help` documentation using `/` for helpful information
- [x] I have read the [frequently asked questions](https://github.com/Jelmerro/Vieb/blob/master/FAQ.md)
- [x] I did not find any similar issues in the [issue tracker](https://github.com/Jelmerro/Vieb/issues)

**Addition or change**
Search engines defined in `search` or `searchwords` only support one `%s` that is replaced by the "search string".
Multiple `%s` are currently replaced by the same search string. This makes search engines like:
```
searchwords+=xy~www.foo.bar/?category1=%s?category2=different_%s?category3 ...
```
impossible. It would be nice if the `%s` syntax were expanded to allow multiple query strings.
Something like (`%s`, `%s2`, `%s3`, ... ) would make the previously mentioned search engine possible. 

I'm not sure how you would delimit the different queries. Double quotes (`:open searchengine "query 1" "query 2" ...`) would be fine with me, but I'm open to suggestions.

**Alternatives considered**
You can make separate search engines for common queries for query 1 and use a single `%s` to search for query 2, but queries are almost never predictable enough for this to work well.

yosoymau:
fixes #326 

This enables the user to create custom searchwords that contain multiple "%s".
Example:
You would set this up using:
`:set searchwords+=ex~example.com/%s/%s/%s`
And then you could type the keywords, followed by fields separated by commas into the address bar in the explore mode:
`ex one, two, three`
And that would get replaced with:
`www.example.com/one/two/three`


Jelmerro:
First of all, thanks for the PR, this is actually much easier than I expected, and I like the use of the `while` for it. I just have one question/suggestion for this.

What happens if someone wants to search for something containing commas? I get that this won't be possible if using that as a separator (which is fine), but I think this doesn't check if there are multiple `%s` templates in the string before substituting them. Maybe we should split on `,` and fill as much as possible, then join the remaining parts with `,` again and use them for the last `%s` in a search. This way searches with only one `%s` will keep working as before, and it will be possible to have the final part of a search contain `,` even if splitting on them.

I hope the above makes sense, and if it did, maybe you could also write the above on the help page in the searchwords section, hopefully in clearer language than I described it.

yosoymau:
I agree with you, i am working on it right now. I'm still unsure about how to handle the case in which the user doesn't type enough queries to fill all the instances of `%s`. The current behavior is that the string gets replaced to undefined and i don't think that this is good enough.

Any suggestions would be appreciated.

Jelmerro:
Glad you will continue working on this.

I agree that undefined is not nice, so I would suggest an empty string as a fallback for those cases.

app/util.js
```diff
@@ -109,12 +109,19 @@ const searchword = location => {
...
                 const queries = q.split(",").map(item => item.trim())
-                let i = 0
                 let urlString = url
-                while (urlString.includes("%s")) {
-                    urlString = urlString.replace(/%s/, encodeURIComponent(queries[i]))
-                    i += 1
+                let counter = 1
+                const patternMatches = (urlString.match(/%s/g) || []).length
+                while (urlString.includes("%s") && counter < patternMatches) {
+                    if (0 in queries) {
```
Jelmerro, line 116: Is this the same as `if (queries[0]) {` if so, I would prefer that syntax for consistency with the rest of Vieb's code.

app/util.js
```diff
@@ -109,12 +109,19 @@ const searchword = location => {
...
+                        urlString = urlString.replace(/%s/, encodeURIComponent(queries.shift()))
+                    } else {
+                        urlString = urlString.replace(/%s/g, "")
+                    }
+                    counter += 1
                 }
+                const remainderString = queries.join(", ")
```
Jelmerro, line 123: If you are splitting on `,` I also expect them to be joined with `,` and not with `, `.

app/util.js
```diff
@@ -109,12 +109,19 @@ const searchword = location => {
             const q = location.replace(`${word} `, "")
             if (q && location.startsWith(`${word} `)) {
                 const queries = q.split(",").map(item => item.trim())
```
Jelmerro, line 111: I don't think we specifically need a trim here actually, as the remainder might have spaces and commas mixed that users would expect to keep and be inserted in the final search part. Maybe add a note to the help about how the remainder works and that spaces are kept.

Jelmerro (commented):
This is a lot better, works pretty neat. Just some final things and I think we will be good to go.

yosoymau:
Thanks for the reviews and the patience! This is my first time contributing to a project and i am learning a lot just looking at the code and hearing your suggestions. I will fix this and send more commits later. Thanks again!

app/util.js
```diff
@@ -108,19 +108,19 @@ const searchword = location => {
...
-                const queries = q.split(",").map(item => item.trim())
+                const queries = q.split(",")
                 let urlString = url
                 let counter = 1
                 const patternMatches = (urlString.match(/%s/g) || []).length
                 while (urlString.includes("%s") && counter < patternMatches) {
-                    if (0 in queries) {
-                        urlString = urlString.replace(/%s/, encodeURIComponent(queries.shift()))
+                    if (queries[0]) {
+                        urlString = urlString.replace(/%s/, encodeURIComponent(queries.shift().trim()))
```
yosoymau, line 117: This is trimmed here to allow for this kind of use case:
`ghrepo Jelmerro, Vieb`
This will get changed to:
`github.com/Jelmerro/Vieb/`

app/util.js
```diff
@@ -108,19 +108,19 @@ const searchword = location => {
...
                     } else {
                         urlString = urlString.replace(/%s/g, "")
                     }
                     counter += 1
                 }
-                const remainderString = queries.join(", ")
+                const remainderString = queries.join(",").trim()
```
yosoymau, line 123: This gets trimmed here to avoid this behavior:
`ghrepo Jelmerro, Vieb`
being replaced to:
`github.com/Jelmerro/ Vieb`

Side effect is that the case in which there's only one %s will also loose the initial spaces...

I'd be ok with that but maybe someone really needs their spaces at the begginning of their searches...

Jelmerro:
You're welcome, glad to have this feature in Vieb now :slightly_smiling_face:

PR #332, opened 2021-11-13
Created setting to filter follow mode elements

### LINKED ISSUE #323: Limit follow mode to a set of elements
**Checklist**
- [x] I searched the `:help` documentation using `/` for helpful information
- [x] I have read the [frequently asked questions](https://github.com/Jelmerro/Vieb/blob/master/FAQ.md)
- [x] I did not find any similar issues in the [issue tracker](https://github.com/Jelmerro/Vieb/issues)

**Addition or change**
Some way to limit the types of elements that are shown in follow mode. I was thinking this could be a setting that takes in a list of element type names, e.g. `:set followElements=text,clickable,links`. The element types would include any of those mentioned in `:h action.startFollowCurrentTab`. 

**Alternatives considered**
I can tell element types apart from the border color of the link hints, though I would prefer if I could make it so the types of elements that I don't want to use are not an option.

yosoymau:
fixes #323 
## Changes:
* Added a new setting (followelement) to allow filtering of the highlighted elements on follow mode.
* Updated the help page to explain the new feature.

### followelement setting
* The types are those mentioned on `:h action.startFollowCurrentTab`.
* Any combination of elements is allowed.
* The setting can be used like: `:set followelement=url,onclick,inputs-insert,inputs-click`


app/pages/help.html
```diff
@@ -678,6 +678,8 @@ <h3 id="firefoxmode">firefoxmode</h3>
...
         <h3 id="fontsize">fontsize</h3>
         Change the font size for the entire Vieb interface. Changing the font size setting will also scale the navbar, built-in pages such as this help page, notifications and <a class="follow" href="#action.startFollowCurrentTab">follow mode</a> indicators. It will not automatically zoom websites, which can be done with <a href="#action.zoomIn">action.zoomIn</a> and related actions.
+        <h3 id="followelement">followelement</h3>
+        This setting can be used to filter the types of elements that will be highlighted when using the <a href="#action.startFollowCurrentTab">follow mode</a>. Different sets of elements can be selected at any moment to be highlighted. For example, with this setting set on "inputs-click,inputs-insert", only input elements will be highlighted in follow mode. Please check <a href="#action.startFollowCurrentTab">follow mode</a> for more information about the types of elements.
```
Jelmerro, line 682: Can you give the `a` tags a `class="follow"` to highlight them in the follow mode color? Thanks!

app/renderer/settings.js
```diff
@@ -93,6 +93,7 @@ const defaultSettings = {
     "favicons": "session",
     "favoritepages": "",
     "firefoxmode": "never",
+    "followelement": "url,onclick,inputs-insert,inputs-click,other",
```
Jelmerro, line 96: This line is all good, but as this is a list of values that has specific values being valid for it, could you check for validity for values entered with `:set` by doing something similar as is done for modifiers on line 416 for example?

Jelmerro (commented):
This is nice,, very simple and clean implementation. Just some small things to make it complete.

Jelmerro (approved):

Jelmerro:
Thanks for implementing this feature!

PR #345, opened 2022-01-09
update cheatsheet

ottopiramuthu:
Just made some edits to the commands, including adding `runjsinpage`. How would you want to include information about command ranges, if at all?


Jelmerro:
How about an example or two in the tab management or split sections that use ranges, and then redirect them to `:h ranges` for more info on it?

I also added new version of follow mode, maybe we can expand the follow mode section with other types of follow mode (we new have "current", "newtab", "split" and "vsplit"). See [this commit for details](https://github.com/Jelmerro/Vieb/commit/f4f7e5aceb0740de82c64531b6f6d7ba66d40120).

I have also just added two new types of link-related pointer mode actions: "splitVideo", "splitAudio" etc. and "vsplitVideo", "vsplitAudio" etc, so it would be great if these could also be added on the bottom right to the table, as well als two new mappings in the visual mode section: "tS" and "tV" which do exactly what you would expect: open the text in a new split and vsplit respectively, see [this commit for all of them if needed](https://github.com/Jelmerro/Vieb/commit/e8d4bfd1d6cc9fb7c485554f061831764dc4e886).

Sorry for making you work overtime on this cheatsheet, but I can't seem to figure out how to edit it without messing up the layout completely :slightly_frowning_face:, you are also using Inkscape right?.

ottopiramuthu:
I don't mind the work, the cheatsheet should be as helpful as possible :) I am using Inkscape.

ottopiramuthu:
Let me know if the latest commit is what you were looking for.

Jelmerro:
The additions to the SVG look nice, but the PNG isn't exported correctly, it looks like you only exported one section: https://github.com/Jelmerro/Vieb/pull/345/files#diff-35e85608ff15e384ac0fbb5f3a0e2d3c8f1268040a3fc0e55da27a7a6ca7cdbfR3077

ottopiramuthu:
Sorry about that. It should be fixed now.

Jelmerro:
Nice, this looks really good, just wondering if/where you added the new follow modes. Do you think we can do without them in the cheatsheet? They might also not be easily list-able in the current key listing format.

ottopiramuthu:
I've added the follow modes in the keyboard under `S` and `V`. In the current format the other thing I can do is add `/S/V` to the `f/F` in the follow mode title.

Jelmerro:
Yes, I meant that they could also be added to the follow section as `/S/V` indeed. Maybe an additional line that follow mode can also open links in new tabs or splits would be nice, so people understand why there are four different keys for one mode.

I also noticed another small thing just now when looking at it, "S" is mapped to "follow mode split", this is listed correctly, but there is no mapping listed for "s", which actually does exist, and does the same as when in pointer mode (it moves the pointer to the mouse position, and also starts pointer mode).

I think we are good to go after these two changes :tada:

ottopiramuthu:
I've made your suggested changes in the latest commit.

Jelmerro:
Thanks again for keeping this up to date!

PR #353, opened 2022-01-30
update cheatsheet

ottopiramuthu:


Jelmerro:
Thank you for the fast update once again! This is unrelated to what was just added (which is nice), but for some reason I never noticed that `I` and `O` are missing from the cheatsheet. Could these be added as well (if you have time for this :sweat_smile:)? `I` moves the tab all the way to the start, `O` all the way to the end (see `:h O` and `:h I` for more info).

ottopiramuthu:
I've added the mappings in the latest commit.

Jelmerro:
Thanks so much!

PR #363, opened 2022-03-31
Small Download bug fix

cab-1729:
Slight deviation from how downloads work in other browsers.

This disrupts the filenames in case of extensions with two dots.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -797,7 +797,10 @@ ipcMain.on("create-session", (_, name, adblock, cache) => {
         let newFilename = item.getFilename()
         while (isFile(save)) {
             duplicateNumber += 1
-            const extStart = filename.indexOf(".")
+            let extStart = filename.lastIndexOf(".tar.")
+			if (extStart === -1) {
+				extStart = filename.lastIndexOf(".")
+			}
             if (extStart === -1) {
                 newFilename = `${filename} (${duplicateNumber})`
             } else {
--- a/app/index.js
+++ b/app/index.js
@@ -793,11 +793,11 @@ ipcMain.on("create-session", (_, name, adblock, cache) => {
         }
         const filename = item.getFilename()
         let save = joinPath(downloadSettings.downloadpath, filename)
-        let duplicateNumber = 1
+        let duplicateNumber = 0
         let newFilename = item.getFilename()
         while (isFile(save)) {
             duplicateNumber += 1
-            const extStart = filename.lastIndexOf(".")
+            const extStart = filename.indexOf(".")
             if (extStart === -1) {
                 newFilename = `${filename} (${duplicateNumber})`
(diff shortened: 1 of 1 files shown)


app/index.js
```diff
@@ -793,11 +793,11 @@ ipcMain.on("create-session", (_, name, adblock, cache) => {
         }
         const filename = item.getFilename()
         let save = joinPath(downloadSettings.downloadpath, filename)
-        let duplicateNumber = 1
+        let duplicateNumber = 0
```
Jelmerro, line 796: This is a good change, thanks for noticing.

app/index.js
```diff
@@ -793,11 +793,11 @@ ipcMain.on("create-session", (_, name, adblock, cache) => {
...
         let newFilename = item.getFilename()
         while (isFile(save)) {
             duplicateNumber += 1
-            const extStart = filename.lastIndexOf(".")
+            const extStart = filename.indexOf(".")
```
Jelmerro, line 800: This does not always have the right effect either, for example:
image
image

I propose that we add an exception for `tar.something` files since those are probably the only cases where `lastIndexOf` is not working right currently. I'm not a fan of using a regular `indexOf`, because when there are dots in the filename it's going to be messed up. What are your thoughts?

Jelmerro (commented):
Thanks for this PR, we can definitely improve the file naming of downloads in Vieb, but I don't think your current implementation is an improvement for all cases. Please see my comments below for details.

veikolippand:
Working solution can be achieved with regular expressions with list of known exceptions where dot is a part of extension.
And i'm not fan of this pattern what most of the popular browsers are using to create different file names. Symbols like "(", ")", " " are not always terminal friendly. Maybe there is a shell what can handle spaces and special characters in file names without escaping but i have not met any yet. So it would be much more convenient if browsers add no special characters but underscores with numbers to create new file names. "cat_pic.jpg" to "cat_pic__2__.jpg" or similar.
And yes i use 100% terminal shell to manage files which might be my personal problem but nevertheless.

cab-1729:
I would agree with having an exception for ```*.tar.*```. Checking for all possible extensions would be too cumbersome. I wonder how the other browsers handle this.
It's better to stick to the standards. Even though parentheses don't work as smoothly in the shell.

veikolippand:
yes you can make check list or regexp only for these extensions you know. But there is nothing you can do with files like _filename.my.awesome.extension_. And these cases you take last part of it until this extension is popular enough to be added extensions list. And i think local MIME database can be helpful to include to identify exotic file extensions what some user application might generate.

cab-1729:
I've added a possible solution for tar files.

Jelmerro (approved):
This works perfect, thanks!

PR #374, opened 2022-05-07
added new theme 'compact.css'

stevkazt:
Hey, thanks for working on this browser I'd like to contribute and I think a good start is a theme, I made it few days ago for me and I think someone else may like it, I moved navbar and tabs to the same line so you can get even more space for page view and focus on keybindings usage, icon and mode space are smaller.

[Check a screenshot here](https://i.imgur.com/31AWR1Q.png)

### DIFF
--- a/app/colors/compact.css
+++ b/app/colors/compact.css
@@ -7,12 +7,8 @@
  * APP
 */
 /* navbar & tabs */
-#navbar {width: 40vw; height: 1.8em;}
-#logo {height: 1.2em;width: 1.2em;min-width: 1.2em;min-height: 1.2em; padding:0.2em;}
-#mode-container {width: 6em;}
-#mode {font-size: 1.1em;}
-
-#tabs {width: 60vw;height:1.8em;}
+#navbar {width: 40vw}
+#tabs {width: 60vw}
 
 /* pages container */
 #page-container{flex: 0 1 auto;}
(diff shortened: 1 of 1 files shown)


app/colors/compact.css
```diff
@@ -0,0 +1,18 @@
...
+
+/*
+ * APP
+*/
+/* navbar & tabs */
+#navbar {width: 40vw; height: 1.8em;}
+#logo {height: 1.2em;width: 1.2em;min-width: 1.2em;min-height: 1.2em; padding:0.2em;}
+#mode-container {width: 6em;}
+#mode {font-size: 1.1em;}
+
```
Jelmerro, line 14: I don't really think this creates extra space, it only changes the sizes slightly, but the margins remain. Removing lines 11-13 will make the theme simpler, and easier to understand, while at the same time maintaining the default look of Vieb better. What do you think?

Jelmerro (commented):
I like this idea of a single bar combining the navbar and the tabs, it looks pretty good.

app/colors/compact.css
stevkazt: Perfect I just did a couple of changes to keep the default look, it is still good for me and happy if we can share it.

Jelmerro: Thanks, looks good now!

PR #384, opened 2022-06-22
reduces page height on compact theme to fit the page view

stevkazt:
For some reason the bottom of the page goes outside of the viewport. I added a line as a quick fix but I am not sure if it is the correct way to solve it.

Check the bottom and the scroll bar to notice the difference: 

[Error](https://i.imgur.com/dmFxk7J.png)
[Fix](https://i.imgur.com/gl6Yd9l.png)


Jelmerro:
I already noticed this, and believe it's already fixed with the current code, but it's not in a stable release yet. Could you copy the current existing code to a new theme in `~/.vieb/colors` or a `colors` folder in your datafolder to test this on your computer?

PR #391, opened 2022-07-01
Vieb bookmarks 

yosoymau:
#100


yosoymau:
This is what i have right now.
Now things are less convoluted than before as i moved everything to a new file, more organized. I have some things i'd like to discuss. 

## The `bmadd` command
What should we expect `bmadd` to do and how? Right now i have it set up like this:
### Case 1:
`bmadd`
If the user sends the command with no arguments it creates a new bookmark, with the current page's url, title and name.
### Case 2:
`bmadd github my repos`
When the user gives the command just a string of text, that string is considered to be the name, and the url and title are set to the current page's data.
### Case 3:
`bmadd name=github~title=main page of github~url=https://github.com`
The user can give to the command a string like this, using this kind of style: `option=value`, and every option separated by a `~`.

My concerns are:
1. Should the bookmarks be filled automatically if some piece of data is missing? 
2. In case 3, should the missing data be filled automatically from the current page? Maybe this could be a setting?
3. My mind tells me that the url is kind of essential to the concept of what a bookmark even is, so i think this should be filled automatically if the user fails to specify one, but should any property of a bookmark be required? 
4. If the user doesn't give a name for a bookmark? What should be the result? Right now it just gets the current page's title, but this seems bad to me.
5. I think the formatting of the string is not a very good one because both equal signs and tildes can be used in urls and titles, and that would mess things up, but i don't know what could be appropriate.
6. Also for the formatting, how should we accept list like properties? if the user wanted to give multiple tags to a bookmark, what would be a good format? Right now these are not supported.

Sorry if this are too many questions.

app/renderer/bookmarks.js
```diff
@@ -0,0 +1,128 @@
...
+    "name",
+    "path",
+    "tag",
+    "title",
+    "url"
+]
+
+const setBookmarkSettings = () => {
+    const setFile = getSetting("bookmarksfile")
+    if (setFile === "bookmarks") {
```
Jelmerro, line 49: Also make this line match for empty (or trimmed empty) setFile values.

app/renderer/bookmarks.js
```diff
@@ -0,0 +1,128 @@
...
+    // Fill missing essential data from relevant sources
+    if (typeof newbookmark.name === "undefined") {
+        newbookmark.name = currentTab().querySelector("span").textContent
+    }
+    if (typeof newbookmark.title === "undefined") {
+        newbookmark.title = currentTab().querySelector("span").textContent
+    }
+    if (typeof newbookmark.url === "undefined") {
+        newbookmark.url = currentPage().src
+    }
```
Jelmerro, line 90: Can you make all string comparisons here rely on the falsyness/truthyness of empty/filled strings? as in:
```js
    if (!newbookmark.name?.trim()) {
        newbookmark.name = currentTab().querySelector("span").textContent
    }
```

app/renderer/settings.js
```diff
@@ -187,6 +189,7 @@ const defaultSettings = {
     "splitright": false,
     "startuppages": "",
     "storenewvisits": "pages",
+    "suggestbookmarks": "all",
```
Jelmerro, line 192: Should be part of `suggestorder`, please see my later comments in #100, or let me know if you have any questions on this. Ideally there should "only" be a `bookmarksfile` setting to and the file contains all the bookmarks, and how they are suggested is controlled by `suggestorder` like other suggestions.

app/renderer/settings.js
```diff
@@ -228,6 +235,7 @@ const listLike = [
     "spelllang",
     "startuppages",
     "storenewvisits",
+    "suggestbookmarks",
```
Jelmerro, line 238: As such, this can go.

app/renderer/settings.js
```diff
@@ -46,6 +46,7 @@ const {
     updateGuiVisibility,
     getMouseConf
 } = require("./common")
+const {setBookmarkSettings} = require("./bookmarks")
```
Jelmerro, line 49: Please import this as needed when used, please only import `common`, `util` and packages at the top (outside of very specific cases). This is faster and uses less memory when bookmarks are not accessed.

app/renderer/settings.js
```diff
@@ -315,6 +323,10 @@ const numberRanges = {
     "timeoutlen": [0, 9000000000000000]
 }
 let customStyling = ""
+const bookmarkSettings = [
```
Jelmerro, line 326: You don't need this list, just check for the `bookmarksfile` setting and call the function directly.

app/renderer/settings.js
```diff
@@ -891,6 +904,10 @@ const updateDownloadSettings = () => {
     ipcRenderer.send("set-download-settings", downloads)
 }
 
+const updateBookmarkSettings = () => {
+    setBookmarkSettings()
```
Jelmerro, line 908: Import the function from `bookmarks` here instead.

app/renderer/settings.js
```diff
@@ -1113,6 +1131,9 @@ const set = (setting, value) => {
         if (downloadSettings.includes(setting)) {
             updateDownloadSettings()
         }
+        if (bookmarkSettings.includes(setting)) {
```
Jelmerro, line 1134: This can just directly check for `boomarksfile`, as the other setting can go, and because `suggestorder` updates don't mean that an update is needed in bookmarks.

Jelmerro (commented):
Thanks so much for starting work on this, looks fairly good, just a lot of small nitpick things :wink:.  Your overall approach and setup seem solid and correct. As for your other questions:

1. Yes, all info should have a fallback, even if empty in some cases. Pretty much everything besides the url should be considered optional in a bookmark.
2. Yes, if users don't want the info they can delete it later I think, unless you think a setting is needed as well.
3. Only the url is required, so if there is no page url, we need the user to supply one (only for the newtab page actually, but other pages are all fair as the url fallback).
4. Page title is fine to me, users can edit this later or set the title they want to.
5. I would imagine adding a url field specifically to the bmadd command not being that common, as usually users would like to simply add the current page, not some other url. Having said that, it should still be supported. I'm also not that big of a fan of how the command currently looks with lots of arguments, initially I thought of automatically guessing what each argument of the bmadd command would be for, but with so many different fields/options to a bookmark that seems not possible, so we do need this actually. I think the tilde is fine, it needs to be escaped in proper urls anyway, and is not stored as such in the bookmarks file, so even if users can't add it with the command when they are not on the page, they can still add it when they ARE on the page or by editing the bookmarksfile. Equal signs should be allowed in the options, you can split each part on the `=` and take the first part as the name, then join the other parts with `=` to get the remaining string with `=` intact.
6. Commas seem like a good option to me. Ideally the tags should follow simpler naming rules, you can use the specialchars variable of `util.js` to check for characters that should be blocked. Maybe choose the one that does allow spaces.

Hopefully you can continue with all these comments and answers, let me know if anything is still unclear. Finally please also refer to my later comments in #100 for a general description of how it should work.

app/renderer/settings.js
yosoymau: I think this is what you mentioned in #100 that the setting `suggestbookmarks` was going to refer to which bookmarks should be included in the suggestions. Maybe we could use this to specify only bookmarks containing a certain tag to be suggested, and others only be accessible through `:bmload`. 
Other option could be to add it as a flag on the bookmark itself.

Jelmerro: Sorry for that, yes you are right. `suggestorder` is only for if/where/how many suggestions should show, `suggestbookmarks` is for optionally filtering which tags should be listed (or default to all). My comment about `suggestbookmarks` not affecting anything that happens in the bookmarksfile still stands, and as such you don't need to write anything to the file after changing the value.

yosoymau:
So now i've pushed an update where i started to add suggestions to the bookmark commands, it's still not working properly, i'm still not sure what's wrong, but i believe it's on the right track.  Thanks for the feedback!

yosoymau:
Hello @Jelmerro good day!
I've been thinking about the way that the command `:bmload` should work. I was working on the suggestions under the assumption that the user would load a bookmark referring to it by it's name, but then i remembered about the keywords and tags, and i'm not sure how these would fit here.

Can you explain what was the purpose that you intended for keywords, tags and names or how you expected them to work when you made the specifications on #100? This is so i can better understand what i should be aiming for, because right now i feel like i'm just kinda using the name as something like a keyword.

Also what do you think of this idea:
1. The user can do `:bmload mytag`and all the bookmarks with that tag get loaded.

Thanks!

app/renderer/bookmarks.js
```diff
@@ -0,0 +1,157 @@
...
+        newbookmark.title = currentTab().querySelector("span").textContent
+    }
+    if (!newbookmark.url?.trim()) {
+        newbookmark.url = currentPage().src
+    }
+    if (!newbookmark.path?.trim()) {
+        newbookmark.path = "/"
+    }
+    if (!newbookmark.bg?.trim()) {
+        newbookmark.bg = "some default hex value"
```
Jelmerro, line 105: You can use the theme colors as the defaults, for example:
https://github.com/Jelmerro/Vieb/blob/master/app/renderer/settings.js#L958

app/renderer/settings.js
Jelmerro: Can you remove "all" from here and just make it a list of tags to filter? And if empty, just show all?

Jelmerro:
The tags and keywords should also be searchable using bmload, probably using the same syntax as bmadd, but i like the idea of just mentioning a tag and loading all of them. Not sure if the same should happen for keywords, but for tags that would be nice. The suggestions for bmload should help letting users know which bookmarks match their query, and load them all or just the one if selected.

The big difference of tags and keywords, should be that tags can be colored and have their own set of keywords, see https://github.com/Jelmerro/Vieb/issues/100#issuecomment-1147360747. This should be all the more useful once there is a bookmarks page to visualize the bookmarks in the folders, where they are shown with these tags in their own colors.

app/renderer/bookmarks.js
yosoymau: Ok, but now thinking about it, if someone changes their theme, they're going to end up with some hard coded themed values in every bookmark. 
Those could be changed later, but i don't know, it seems bad to me. 
Maybe we could make it so this option accepts something like bg-default and fg-default, that then resolve to the defaults of the theme while rendering or something like that. What do you think?

Jelmerro: That's a valid concern actually, maybe we should not add the colors to every entry, and just fallback to the theme colors in case they are missing when it actually happens, not when making them.

app/renderer/settings.js
yosoymau: Done!

app/renderer/bookmarks.js
yosoymau: For now i've just removed the auto filling of those values.

yosoymau:
I'm starting to work on the special page to manage bookmarks, and i ran into a roadblock that i haven't been able to solve. I'm trying to get the bookmark data from a function in `renderer/bookmarks.js` to `preload/bookmarks.js`. Now i found out these exist in different contexts, as far as i understand, these can't communicate directly. Is this right?

So my guess is that this should be done via `ipcRenderer` messages. I was looking around the code i dont have any idea on how i could get the data from `renderer/bookmarks.js` to maybe `index.js` (? not sure about this) and then get that sent to `preload/bookmarks.js`.

Any suggestion would be appreciated.

Jelmerro:
You are correct, I would like to recommend the newtab page as a quick and easy example on how to do ipc-renderer communication: https://github.com/Jelmerro/Vieb/blob/master/app/preload/newtab.js, you can find the other side of things in a range of places, but the first point of action is here: https://github.com/Jelmerro/Vieb/blob/6d06e23ca71af39269f1b1fcf46df758ce8129af/app/renderer/tabs.js#L867. You do NOT need to go through main (unlike a lot of other things), as you don't need to interact with subframes or main-only APIs.

Basically, from renderer to preload/pages you can do `webview.send` and inside the preload you can do `    ipcRenderer.sendToHost` to send it back to the renderer, which is received in the tabs file as an event on the webview. If you do `ipcRenderer.send` from the preload, you end up in main, but you won't need that in this case.

yosoymau:
This was exactly what i needed, i'll be back when i get something working. Thanks a lot.

yosoymau:
Now i've added the start of the special page for managing bookmarks, and this is what i got. 
I don't know if this is the direction we were aiming for, right now it's kinda ugly, because it needs some css love, and things like tags or the editing functionality are not implemented yet. 

My question is this: is this kind of solution good?

Jelmerro:
Hey @yosoymau I hope you are well, I was wondering if you were planning to continue work on this PR. Thanks for the work so far, let me know if you need help with anything!

yosoymau:
Hello again, i'm back and i'll keep working on this, so expect to hear back from me soon.

I apologize for my absence and leaving this halfway for such a long time.

Jelmerro:
Welcome back, glad you are willing to continue work on it, just let me know if there's anything you need help with. I'm a bit inactive for the last weeks as well, I will be back in a few.

app/preload/bookmarks.js
```diff
@@ -0,0 +1,135 @@
...
+
+// Create tree structure to keep track of added folders.
+const tree = {
+    "children": [],
+    "folderName": "/",
+    "fullpath": "/"
+}
+
+ipcRenderer.on("bookmark-data-response", (_, bookmarkData) => {
+    // Create folder structure
```
Jelmerro, line 29: This comment seems redundant.

app/preload/bookmarks.js
```diff
@@ -0,0 +1,135 @@
...
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+"use strict"
+const {ipcRenderer} = require("electron")
+
+// Create tree structure to keep track of added folders.
```
Jelmerro, line 21: This comment seems redundant.

app/colors/default.css
```diff
@@ -350,6 +350,9 @@ body:not(#app) ::placeholder {color: var(--placeholder-text);}
 #downloadspage .misc {display: flex;margin: .5em;flex-direction: column;width: calc(100% - 1em);}
 #downloadspage .filelocation {cursor: pointer;}
 #downloadspage img {cursor: pointer;}
+/* Bookmarks  */
```
Jelmerro, line 353: Bookmarks in lowercase here please.

app/preload/bookmarks.js
```diff
@@ -0,0 +1,135 @@
...
+    treeRootList.appendChild(treeRootDetails)
+    treeRootElement.appendChild(treeRootList)
+
+    document.getElementById("bookmarks").appendChild(treeRootElement)
+    update()
+})
+
+const update = () => {
+    ipcRenderer.sendToHost("bookmark-data-request")
+}
```
Jelmerro, line 57:
```suggestion
const update = () => ipcRenderer.sendToHost("bookmark-data-request")
```

app/preload/bookmarks.js
```diff
@@ -0,0 +1,135 @@
...
+    const {folders} = bookmarkData
+    for (let i = 0; i < folders.length; i++) {
+        const folderElements = folders[i].path.split("/")
+            .filter(e => e)
+
+        // Set the tree root as the starting point.
+        let currentCheckingFolder = tree
+        folderElements.forEach(e => {
+            // Ignore / because it already exists.
+            if (e !== "/") {
```
Jelmerro, line 88: Maybe filter these before doing the forEach? And possibly make this into a function as this is intended quite a bit.

app/renderer/bookmarks.js
```diff
@@ -0,0 +1,323 @@
+/*
+* Vieb - Vim Inspired Electron Browser
+* Copyright (C) 20
```
Jelmerro, line 3: This seems incorrect, could you remove the extra newlines and make sure to update the year if you update the file? There are plugins for most editors to do this automatically too, here's one for Vim: https://github.com/antoyo/vim-licenses

app/renderer/bookmarks.js
```diff
@@ -0,0 +1,323 @@
...
+        notify(`Loaded ${selectedBookmarks.length} bookmarks \
+                that matched the query.`)
+        selectedBookmarks.forEach(e => addTab({
+            "url": e.url
+        }))
+    }
+}
+
+const matchBookmarksToInput = input => {
+    const storedBookmarkData = getBookmarkData().bookmarks
```
Jelmerro, line 177: No need for calling the function in this same file, just access `bookmarkData` directly.

app/renderer/bookmarks.js
```diff
@@ -0,0 +1,323 @@
...
+
+const addFolder = path => {
+    const currentStoredPaths = bookmarkData.folders.map(f => f.path)
+    if (!currentStoredPaths.includes(path)) {
+        bookmarkData.folders.push(
+            {
+                "keywords": [],
+                "name": "",
+                path
+            })
```
Jelmerro, line 269:
```suggestion
            {"keywords": [], "name": "", path})
```

app/renderer/bookmarks.js
```diff
@@ -0,0 +1,323 @@
...
+    const currentTagIds = bookmarkData.tags.map(tag => tag.id)
+    tags.forEach(t => {
+        if (!currentTagIds.includes(t)) {
+            bookmarkData.tags.push(
+                {
+                    "id": t,
+                    "keywords": [],
+                    "name": ""
+                }
+            )
```
Jelmerro, line 256:
```suggestion
                {"id": t, "keywords": [], "name": ""})
```

app/renderer/bookmarks.js
```diff
@@ -0,0 +1,323 @@
...
+        notify(`A bookmark with name: ${bookmark.name} already exists.`)
+    }
+
+    addTags(bookmark.tag)
+    addFolder(bookmark.path)
+
+    // Check path format
+    // Check color hex values
+    // Check keywords and tags.
+    // Single words?
```
Jelmerro, line 307: I presume these are still on the TODO list? Let me know if you are stuck and/or need help to implement this :)

app/renderer/bookmarks.js
```diff
@@ -0,0 +1,323 @@
...
+
+    if (badOptions.length !== 0) {
```
Jelmerro, line 309:
```suggestion
    if (badOptions.length) {
```

app/renderer/suggest.js
```diff
@@ -741,6 +741,108 @@ const suggestCommand = searchStr => {
             }
         }
     }
+    [
+        "bmload",
+        "bmdel"
+    ].forEach(
```
Jelmerro, line 747:
```suggestion
    ["bmload", "bmdel"].forEach(
```

app/renderer/suggest.js
```diff
@@ -741,6 +741,108 @@ const suggestCommand = searchStr => {
...
+        bmCommand => {
+            if (bmCommand.startsWith(command) && !confirm) {
+                const {getBookmarkData,
+                    validBookmarkOptions} = require("./bookmarks")
```
Jelmerro, line 751:
```suggestion
                const {
                    getBookmarkData, validBookmarkOptions
                } = require("./bookmarks")
```

Jelmerro (commented):
I went through the current code, most of it looks very good, just some small suggestions so far. I'm very happy with your commitment to build this feature over the past years, let me know if I can help with anything.

TiagoRCorreia:
Still in development?

Jelmerro:
I haven't seen @yosoymau around for a while, so this PR has been stale indeed. I am personally in no rush to get this implemented, but have outlined my proposal in the main ticket #100. I'm not going to rush @yosoymau to do it for us, they can take as long as they need to implement this huge feature, if at all, any contribution is optional and they shouldn't feel obliged to finish the PR, that's why it's still in draft as well. Feel free to pick up where this PR left and make a new one, or please wait for it to be done patiently.

Jelmerro:
Given the lack of recent development, it seems like somebody else decided to continue working on this. Hence I'm closing this in favor of #609, feel free to reopen or create a more up to date one if needed. Thanks for all your work on this.

PR #397, opened 2022-08-04
small links finding optimization in parseAndDisplayLinks

txgk:


Jelmerro:
That does seem better, even if small, thanks a lot!

PR #398, opened 2022-08-07
use regex to detect svg format instead of relying on is-svg

txgk:


Jelmerro:
I first want to thank you for making this PR, as this is something that definitely helped improve Vieb. The reason for closing it, is that I was inspired by this PR to improve the check with a list of file extensions to check for as well as improving the regex you are using (`*` will match the character before it 0-many times, so it should have been `.*` instead :wink:). Thanks for making this, but I'll close it in favor of https://github.com/Jelmerro/Vieb/commit/bc73aec02d8803a8f35d0c89e6ee736ecbd52caf.

txgk:
That's nice. I'm glad that the number of dependencies has decreased. Removing is-svg from the project has decreased the size of resources by more than 100 kilobytes. All 100 kilobytes just for this tiny little check! To be honest I continue to be bothered by the abundance of those little dependencies that go along with the primary dependencies. Relative to projects in other languages, I more often see news about JS package maintainer [atrocities](https://github.com/advisories?query=malicious+package+severity%3Acritical+type%3Areviewed+ecosystem%3Anpm) - this makes me think every time I enter my bank details on some website... I understand that you can't just remove all dependencies, I understand that they are needed for things. Maybe in the case of a not very large dependencies, you can simply copy it to the repository to provide at least a tiny bit more of security.

Jelmerro:
One thing that helps massively, especially if you want to build the app is doing `npm ci --prod` instead of a regular `npm ci`, though I do agree that the amount of micro-dependencies of packages is way too high in some projects. In Vieb I try to minimize them, and also look at the tree of dependencies often to minimize it. The `is-svg` package has only one dependency, an xml parser, which also had only 1 dependency, and the tree ends there. It's around 100kb indeed, but there are far worse offenders in this regard. Still, thanks for bringing this to the attention.

txgk:
> One thing that helps massively, especially if you want to build the app is doing `npm ci --prod` instead of a regular `npm ci`

Yeah, I already got the idea that dev dependencies are not needed in release builds. By the way, npm on my system says that the `--prod` flag is deprecated and `--omit=dev` should be used now.

> I do agree that the amount of micro-dependencies of packages is way too high in some projects. In Vieb I try to minimize them, and also look at the tree of dependencies often to minimize it.

Good to know that this is under spotlight around here :)

> It's around 100kb indeed, but there are far worse offenders in this regard.

Oh that's for sure

PR #403, opened 2022-08-15
add quickmarking to cheatsheet

ottopiramuthu:
Let me know if there are other features you want added from 9.0.0.


Jelmerro:
Het @ottopiramuthu, thanks again for updating this. It would seem that you forgot to update `m/M` and `'/"` in the keyboard itself with these new keys. If you could add them there as well, this can be merged :+1:

ottopiramuthu:
I've added the mappings to the keyboard in the latest commit.

Jelmerro:
Thanks!

Issue #427, opened 2022-10-22
action.openNewTabWithCurrentUrl not filling URL


KunaPrime:
**Checklist**
- [x] I searched the `:help` documentation using `/` for helpful information
- [x] I have read the [frequently asked questions](https://github.com/Jelmerro/Vieb/blob/master/FAQ.md)
- [x] I did not find any similar issues in the [issue tracker](https://github.com/Jelmerro/Vieb/issues)

**Describe the bug**
`nmap T <action.openNewTabWithCurrentUrl>` opening new tab with empty url

**To Reproduce**
1.  go to any web page
2. press `T`
3. new tab opens with URL empty

**Expected behavior**
new tab should open with pre-filled with url current tab 

**Specs**
 - OS: Ubuntu 20.04
 - Version: 9.3.0, Electron 21.1.0 and the Chromium version is 106.0.5249.91.

Jelmerro:
I have made a comment, but in short, this action is not to duplicate a tab, but to open a new empty one, and have explore mode with the current url filled in without navigating directly, so that you can either edit it, or press enter to navigate.

Jelmerro:
Fixed in 9.4.0!

KunaPrime:
i have tried it and it works! thank you

PR #431, opened 2022-10-25
possible fix for #427

KunaPrime:


Jelmerro:
The idea is that you can open a new tab with the current url while in explore mode, not that it opens a new identical tab. For some reason we end up in normal mode since a few releases, but the intended behavior as per docs is to end up in explore mode with the url filled in.

KunaPrime:
i think that problem is in  `setInterval(() => ACTIONS.setFocusCorrectly(), 500)`  namely when new tab is opened main window is in "empty" state so then url falls back to empty one.  I'm not sure how to circumvent this so i did what seemed to be close enough.  

Also can one emulate behavior that i have implemented here with existing actions?

Jelmerro:
This interval has been there since the inception of Vieb, so I am fairly certain it's not the direct cause, but maybe some contents in `setFocusCorrectly` have recently changed that broke this functionality. And yes, your use case is possible if this bug wouldn't exist, since you could map this action with an additional enter key after it.

Jelmerro:
Given the lack of response, I started to look into this myself, and will commit a fix for this shortly. Thanks for taking the time to make a PR!

PR #432, opened 2022-10-29
add startFollowCopyLink action

txgk:
The similar functionality was mentioned in https://github.com/Jelmerro/Vieb/pull/89 but using `vfabcdyy` where `abcd` is the hint of the element in follow mode seemed pretty cumbersome to me. Also such an action cannot be written in the nmap setting, because `nmap yf vfyy<ESC>` will select element with `yy` hint and won't copy anything. In qutebrowser there's a `hint links yank` and for a long time I wanted such functionality in Vieb so here it is :)


Jelmerro (approved):
Good idea to use `yf` as a mapping for it. Very nice PR, no comments from me, thanks for making this!

PR #479, opened 2023-05-16
Fix: cant write to disk erwic settings that match defaultSettings

saltrafael:
`defaultSettings` has for example `containernewtab` set to `s:usecurrent`, and when running an erwic application the setting is set to `s:external`, but when trying to set it back to `s:usecurrent` and running `mkviebrc` to write to disk, the setting can't be saved as it still matches the `defaultSettings` default, so erwic applications will always run overriding these settings that can't be saved. This change instead replaces the `defaultSettings` value when running an erwic application, and then allowing that setting to be written to disk when changed as it won't match the new defaults

### DIFF
--- a/app/renderer/settings.js
+++ b/app/renderer/settings.js
@@ -1299,16 +1299,17 @@ const loadFromDisk = (firstRun = true) => {
         validOptions.adblocker = ["off"]
     }
     if (firstRun) {
+        if (isFile(joinPath(appData(), "erwicmode"))) {
+            defaultSettings.containernewtab = "s:external"
+            defaultSettings.containerstartuppage = "s:usematching"
+            defaultSettings.permissioncamera = "s:allow"
+            defaultSettings.permissionnotifications = "s:allow"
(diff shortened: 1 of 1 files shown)


Jelmerro:
This is not the correct solution, because now the actual Erwic defaults can't be written, even though a datafolder is not erwic specific (a folder might be used for erwic AND regular usage at different times) and because settings are not prefixed with `s:` in all cases, just the special names for the container settings are, others don't need or accept the `s:` prefix.

Jelmerro:
Thanks for making a potential fix for this though, I'll think of a better solution for this or feel free to suggest one!

saltrafael:
Ah, sorry for rushing with this without even opening a tracking issue, just wanted to get this working here so I did the changes, what do you think about the new commits? I think the new erwic defaults const also make it easier to modify in the future by changing one line in the object

Jelmerro:
Yes, this is much better, thanks for fixing this!

Jelmerro (approved):

Issue #385, opened 2022-06-23
Common extension integration in favor of custom installation


Jelmerro:
**Checklist**
- [x] I searched the `:help` documentation using `/` for helpful information
- [x] I have read the [frequently asked questions](https://github.com/Jelmerro/Vieb/blob/master/FAQ.md)
- [x] I did not find any similar issues in the [issue tracker](https://github.com/Jelmerro/Vieb/issues)

**Addition or change**

As had been the case for a few years, extension support in Electron isn't good, neither is the support in Vieb. For this reason, I propose to deprecate and remove the current extension support in favor of integrating commonly requested features into Vieb without the need for an extension. The current list of extensions (or features of said extensions) that are considered are:

- [x] Darkreader (implemented and ready for release)
- [x] Sponsorblock (implemented and ready for release)
- [ ] Password manager, see #106 
- [ ] Greasemonkey or similar, see #64 
- [ ] Stylus or similar (to apply custom styling to a page)
- [ ] Scrollbar customization of some sort (could also be part of stylus-like implementation)
- [ ] Scrolling behavior controlled with [scroll-behavior](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior) and [scroll-snap-type](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-snap-type)
- [ ] Built-in dictionary (as in, show the definition of a word, not just correct misspelled ones)
- [ ] Word replacer (replace random words on the page with different language words with definition, as to learn languages)
- [ ] Page/word translator (related to the two above)
- [ ] Text-to-speech, read pages or section of it (selected with visual) out loud
- [ ] Automatic clearing of cookies, localstorage and cache, based on a timer and/or domain name
- [ ] Bookmarks #100 

All of these should be OPTIONALLY activated, as is already the case for the ones that are marked as done. Please feel free to suggest additional extensions/features you use often and for which there is no existing way to do so in Vieb, either in the comments below or as a new ticket.

**Alternatives considered**

An alternative to the above scenario would be to create and support browser extensions, but please DO NOT underestimate this task. I have tried and failed for years, as have many others. The enormous size of this task is what led to the deprecation of extension support in the first place. If you think you are up for the task, please go ahead: write the code and make a PR for it, but until then, know that building optional features into Vieb will be the current way forward. For info about the previous extension approach, see #130.

yosoymau:
[...]

PR #480, opened 2023-05-19
Add pageRSSLinkToClipboard to copy RSS feed links to clipboard

saltrafael:
a draft attempt at implementing https://github.com/Jelmerro/Vieb/issues/385#issuecomment-1524067991, probably needs a way to select which feed to copy since this gets the first `link` tag and some sites can have multiple per page

peek


Jelmerro:
Thanks for taking the time to implement this! If you could use `includes` instead of `indexOf !== -1`, as that is more in line with the rest of the code.

If multiple RRS feed links are common, maybe a separate action list them and copy a single one would be cool, and copy them by index using a mapping like `yR1`, `yR2` etc. and then `yRL` or `yRR` that could list them. If it's not that common maybe just the first link copy would be good enough.

saltrafael:
Thanks! what do you think now?

https://github.com/Jelmerro/Vieb/assets/76502841/b4b1851c-c057-462c-bdf3-b2be8fe834f5

Jelmerro:
That's very good, I like this implementation a lot. Thank you for taking the time to make this feature, it will land in V10 soon.

PR #491, opened 2023-06-19
use absolute path for RSS feeds

saltrafael:
Turns this:
image
into this:
image
otherwise copies only the path and not the full link

_(example taken from https://invidious.sethforprivacy.com/channel/UCl2mFZoRqjw_ELax4Yisf6w)_


Jelmerro (approved):

Jelmerro:
Thanks for the fix!

PR #497, opened 2023-06-29
Allow usage of current url, current origin as values in commands execution

saltrafael:


app/renderer/actions.js
```diff
@@ -1325,6 +1325,13 @@ const useEnteredData = () => {
     }
 }
 
+const useCurrentUrl = () => getPageUrl()
+
+const useCurrentOrigin = () => {
```
Jelmerro, line 1330: This doesn't feel like the right place to have these templates stored, since they are now also treated as actions. The command replacement logic is good, but this needs moving, probably to `command.js`, also means less imports.

app/renderer/command.js
```diff
@@ -2358,10 +2361,24 @@ const execute = (com, settingsFile = null) => {
     // Remove all redundant spaces
     // Allow commands prefixed with :
     // And return if the command is empty
-    const commandStr = com.replace(/^[\s|:]*/, "").trim().replace(/ +/g, " ")
+    let commandStr = com.replace(/^[\s|:]*/, "").trim().replace(/ +/g, " ")
     if (!commandStr) {
         return
     }
+
+    // Don't "use current" on holdUseCommands, commands like 'command' or 'map'
```
Jelmerro, line 2369: This section is good. Can I suggest that we add more variants of the url? Most importantly domain (origin without protocol, I think it's named `.host` on the url spec). Origin can stay, as that is also useful if you do want the protocol. And just move the replacement functions to this file, they are not actions and shouldn't be treated as such, maybe we have add a paragraph to the command help to explain these templated `use` strings?

Jelmerro (commented):
Thanks for this PR, really cool that we will soon have this. I just have a couple of suggestions to improve the structure, but the replacement logic is mostly good!

PR #505, opened 2023-08-01
Write command for other file types, and using custom path.

yosoymau:
I wanted to have a way to save an image with a custom path so i extended the functionality of the `:write` command so that it allows you to download file types other than html, while letting you set a custom save path for them. I implemented this in a hacky way, but i think it works fine.


yosoymau:
Hello! I've been thinking about this, and i don't think it's a good enough solution. It doesn't support urls that don't include a file extension. Is there a way to access something like `content-type` from the renderer, or to tell if a page is just loading a file? i thought about checking the child nodes of `<body>` and see if there's nothing else other than an image or other file being displayed but the problem there is that i don't know if its possible to know what the correct file extension should be, and also that kind of solution doesn't seem right to me. I haven't been able to come up with other ideas.

Jelmerro (commented):
Hey there, thanks for making this PR. I do agree that in the current state it does not seem very reliable, but the file extension is at least a reliable indicator, so if available I would use that anyway, but as you mentioned it's not there in all cases. For those cases we would need to intercept the headers for the url and check the mime-type. This is possible, but will require further changes in this index file, namely on lines 1039 to 1044, where there is `newSess.webRequest.onHeadersReceived((details, callback) => {` This intercepts the headers and receives the details as an argument. You can ignore/keep the callback, that's merely to indicate if the request should be allowed or not, which is controlled by the adblocker module, but on the details there will be props to indicate the mime type and such, although even that is not 100% reliable, but still better than nothing, so I would use it as a fallback for if there is no url fileytype present. And yes, we will probably need a mapping from mime type to fileype/filename, either via an object or via a module.

PR #506, opened 2023-08-02
Fix media-playing indicator in numberedtabs colorscheme

saltrafael:
the auto margin was causing the media-playing indicator to be hidden. I didn't even notice for a long time that was missing

### DIFF
--- a/app/colors/numberedtabs.css
+++ b/app/colors/numberedtabs.css
@@ -17,5 +17,5 @@
 */
 #tabs, #navbar {counter-reset: tab-counter -1;}
 #tabs > ::before {counter-increment: tab-counter 1;content: counter(tab-counter) ". ";margin: auto 0;}
-#tabs [media-playing]::before {margin: 0 !important}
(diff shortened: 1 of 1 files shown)


app/colors/numberedtabs.css
```diff
@@ -17,4 +17,5 @@
 */
 #tabs, #navbar {counter-reset: tab-counter -1;}
 #tabs > ::before {counter-increment: tab-counter 1;content: counter(tab-counter) ". ";margin: auto 0;}
+#tabs [media-playing]::before {margin: 0 !important}
```
Jelmerro, line 20: Could you be so kind as to add a trailing `;` to this line?

Jelmerro (commented):
Thanks for fixing this, I have no clue how I have never noticed this, but glad it's going to be fixed. Just a tiny change for consistency with other code and we can merge it right away.

Jelmerro (approved):

Jelmerro:
Thank you for fixing this!

PR #549, opened 2024-03-26
Add catppuccin colorscheme

ceuk:
I thought I'd see if you wanted to merge the colorscheme I wrote based around the highly popular [Catppuccin](https://github.com/catppuccin) color palette. 

I think it looks pretty nice, and could make for a nice option for people in addition to the current defaults.

image


amqndin:
that is very cool

app/colors/catppuccin.css
```diff
@@ -0,0 +1,200 @@
...
+[current-mode=command] #mode-container::before,
+[current-mode=insert] #mode-container::before,
+[current-mode=normal] #mode-container::before,
+[current-mode=follow] #mode-container::before,
+[current-mode=explore] #mode-container::before,
+[current-mode=search] #mode-container::before,
+[current-mode=pointer] #mode-container::before,
+[current-mode=visual] #mode-container::before {
+  font-size: 1.5em;
+}
```
Jelmerro, line 171: I think these can be reduced to a single `#mode-container::before` selector instead of targeting each mode explicitly. The `::before` already makes it more specific than any default styling, so it's not needed to list each mode.

app/colors/catppuccin.css
```diff
@@ -0,0 +1,200 @@
+/*
+* Vieb - Vim Inspired Electron Browser
+* Copyright (C) 2019-2023 Jelmer van Arnhem
```
Jelmerro, line 3: It's 2024 by now :)

app/colors/catppuccin.css
```diff
@@ -0,0 +1,200 @@
...
+    --follow-media-border: rgba(166, 227, 161, 0.4);
+    --follow-media-hover: rgba(166, 227, 161, 0.4);
+
+    --follow-image-bg: #f9e2af;
+    --follow-image-border: rgba(249, 226, 175, 0.4);
+    --follow-image-hover: rgba(249, 226, 175, 0.4);
+
+    --follow-other-bg: #cdd6f4;
+    --follow-other-border: rgba(205, 214, 244, 0.4);
+    --follow-other-hover: rgba(205, 214, 244, 0.4);
```
Jelmerro, line 90: Can you use hex colors? You can convert the alpha value as 2 additional digits, making 8 in total. For example, `rgba(205, 214, 244, 0.4)` becomes `#cdd6f466`.

app/colors/catppuccin.css
```diff
@@ -0,0 +1,200 @@
...
+    --syntax-markup-emphasis: #94e2d5;
+    --syntax-markup-addition-fg: #cba6f7;
+    --syntax-markup-addition-bg: #181825;
+    --syntax-markup-deletion-fg: #f2cdcd;
+    --syntax-markup-deletion-bg: #f38ba8;
+/*
+ * APP
+*/
+/* navbar & tabs */
+#tabs, #navbar {width: 100vw;display: flex;background: var(--bg);counter-reset: tab-counter 0;}
```
Jelmerro, line 157: Tab index numbers should start at 0, to be in line with the buffer-like commands. They are now always out of sync with Vieb's commands. To do so, just change the tab-counter to -1.

app/colors/catppuccin.css
```diff
@@ -0,0 +1,200 @@
...
+/* Nav url/search/command */
+#url {display: inline-block;flex: 1;border: solid .1em var(--url-default);margin: .1em;background: transparent;color: var(--fg);outline: none;font: inherit;min-width: 0;padding: 0.5em 0.4em; line-height: 1;width: 0;cursor: default;}
+/* follow */
+#app #follow .follow-url, #app #follow .follow-inputs-click, #app #follow .follow-inputs-insert, #app #follow .follow-onclick, #app #follow .follow-media, #app #follow .follow-image, #app #follow .follow-other {position: absolute;color: var(--follow-text);padding: 0 .1em; border-radius: 0.2em; font-size: 0.8em;}
+/* mouse related */
+#mode-suggestions {background: transparent;max-height: 0;position: absolute;left: 0.5em;top: 2em;padding-top: .1em;z-index: 37;overflow: hidden;display: block;width: 7em;text-transform: capitalize;display: flex;flex-direction: column;}
+#url-hover {display: none;position: absolute;background: var(--url-hover-bg);color:var(--url-hover-fg);bottom: 0;left: 0;margin: 0;pointer-events: none;padding: .1em;white-space: nowrap;overflow: hidden;text-overflow: ellipsis;width: fit-content;max-width: 80vw;text-align: center;z-index: 4;}
+/* hide logo */
+#logo {display: none}
+
```
Jelmerro, line 200: Please delete one of the final newlines at the end of the file, there are now 2, but 1 is enough.

app/colors/catppuccin.css
```diff
@@ -0,0 +1,200 @@
...
+    --follow-click-border:rgba(137, 220, 235, 0.4);
+    --follow-click-hover: rgba(137, 220, 235, 0.4);
+
+    --follow-insert-bg: #eba0ac;
+    --follow-insert-border: rgba(235, 160, 172, 0.4);
+    --follow-insert-hover: rgba(235, 160, 172, 0.4);
+
+    --follow-onclick-bg: #89dceb;
+    --follow-onclick-border:rgba(137, 220, 235, 0.4);
+    --follow-onclick-hover: rgba(137, 220, 235, 0.4);
```
Jelmerro, line 78: The follow mode colors for input elements (click) and JS elements (onclick) is currently exactly the same. I would like these to have different colors by default, so people know what to click on.

Jelmerro (changes requested):
This looks cool, but there are a couple of things I would like to see changed.

- The mode dropdown (the one on hover of the mode indicator) does not align well with the bigger navbar, please move it down a bit, and perhaps also align it to the left if that's preferred. 
- There appears to be no indicator inside explore mode as to what will happen when I click enter. This also means the related help text (see `:h explore`) is now unreadable
image
- Finally, I do not like the required dependency on nerd fonts, this is how it looks for systems without it (which I can imagine are most systems, given users need to install it explicitly on pretty much all distros, let alone other operating systems)
image
I would recommend to replace the font dependency to using CSS data urls, see https://developer.mozilla.org/en-US/docs/Web/CSS/url#using_a_data_url for help.

Overall, I am very happy with this theme, despite all the feedback above. It honestly looks stunning for the most part, and am looking forward to bringing it to Vieb. Please feel free to reach out if you have followup questions regarding the feedback, I should be faster to respond now that I'm back from my holiday.

amqndin:
whoa thats pretty insane, im going to have to update my catppuccin theme, also can you somehow make changing font easier, like use some kind of variable that would change all font family occurences?

ceuk:
These changes all make sense, I have around an hour now so will try get them all done, if not I'll try finish them off soon

ceuk:
Hey @Jelmerro I think that's everything if you want to have another look :+1:

app/colors/catppuccin.css
```diff
@@ -0,0 +1,200 @@
...
+    --syntax-markup-deletion-fg: #f2cdcd;
+    --syntax-markup-deletion-bg: #f38ba8;
+/*
+ * APP
+*/
+/* navbar & tabs */
+#tabs, #navbar {width: 100vw;display: flex;background: var(--bg);counter-reset: tab-counter -1; position: relative;}
+#logo {-webkit-app-region: drag;height: 1.8em;width: 1.8em;margin: .1em;min-width: 1.8em;min-height: 1.8em; display: none;}
+#mode-container {display: flex;width: 4em; height: 4em; text-align: center;margin: 0;justify-content: center; align-items: center;font-size:0.7em; font-weight: normal; position: relative;}
+#mode {text-transform: capitalize;font-size: 1.3em;font-weight: bold;display: flex;align-items: center; line-height: 1;display: none;}
```
Jelmerro, line 159: A lot of these styles overlap with the default styles, they do not need to be repeated. I would also prefer to keep the logo visible if possible.

app/colors/catppuccin.css
```diff
@@ -0,0 +1,200 @@
...
+/*
+ * COLORS
+*/
+:root {
+    /* general */
+    --bg: #11111b;
+    --fg: #cdd6f4;
+    --tab-foreground: #a6adc8;
+    --tab-background: #181825;
+    --tab-suspended: #181825;
```
Jelmerro, line 27: Can we give suspended tabs a different color? I believe this is important information that's not supposed to be gone by using a different colorscheme.

app/colors/catppuccin.css
```diff
@@ -0,0 +1,200 @@
...
+/* tabs */
+#tabs > span {padding: 0.25em;font-size: 0.9em;color: var(--tab-foreground);background: var(--tab-background);display: flex;flex: 1;max-width:20em;overflow: hidden;white-space: nowrap;height: 2.5em; border-top: solid .1em transparent;}
+#tabs > ::before {counter-increment: tab-counter 1;content: counter(tab-counter); font-size: 0.85em;}
+#tabs > span > img {margin: auto 0 auto .25em;padding: 0; padding-right: 0.5em;height: 1em;width: 1em;min-height: 1em;min-width: 1em;pointer-events: none;}
+#tabs .visible-tab {background: var(--visible-tab); color: var(--fg); border-top: .1em solid #74c7ec;}
+#tabs .pinned {min-width: 3em !important;max-width: 3em;width: 3em; padding: 0.25em;}
+#tabs .pinned > span { display: none; }
+#tabs .pinned > img {padding: 0;margin: auto;height: 1.2em;width: 1.2em;min-height: 1.2em;min-width: 1.2em;}
```
Jelmerro, line 183: A lot of this styling is default too, and part of this styling means the suspend color is not being used, so please check with the default colorscheme to remove as much css that's duplicate as possible.

Jelmerro (changes requested):
This is very great, almost ready I think, thank you for the update. I would like to know if it's possible to keep the logo visible. Aside from that, I noticed a lot of duplicate styling compared to the default colorscheme, this is redundant and can be removed. Finally, I would recommend to change the tab-suspended color to `#000000` and make sure it works (as currently it's overwritten by the duplicate styling below). Thank you for working on this, I think  we are very close to merging this, and would like to include this in the next release if possible.

ceuk:
I'll try sort the above later :+1:

artromone:
Any updates? :)

Jelmerro:
I'll be taking this over and make it part of the next release myself as per lack of response from the original author, Vieb 12.5.0 will have light and dark catppuccin themes.

PR #569, opened 2024-10-06
Fix Typo

guideddrone:
On Firefox It's "New Tab"
On Chrome It's "New tab"
in app/pages/newtab.html It's "New tab" so that settles it. P.S: Leads to weird loading behavior where the original name "newtab' is shown without an icon before the page is loaded  I'm on slow hardware so this might not be as big of of a problem as it might seem.


Jelmerro:
thanks for the fix :tada:

PR #610, opened 2025-10-04
[FEATURE] New colorscheme "totalblack"

3mpee3mpee:
new colorscheme for the deep dark colors inspired by [Vanta Black HyDe theme](https://github.com/rishav12s/Vanta-Black) and [Onedark Dark](https://github.com/olimorris/onedarkpro.nvim)

image
image
image


app/colors/totalblack.css
```diff
@@ -0,0 +1,150 @@
...
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+/*
+ * COLORS
```
Jelmerro: Since there are only colors, this comment is redundant.

Jelmerro (commented):
Thanks for taking the time to add this new theme. I think should could be a nice addition, specifically for OLED screens and the like, but there are some small things I would like to have changed. Overall, I think the theme looks really good.

Jelmerro (commented):
Thanks for the update, this looks really great and compact now. Just one more thing I'm a bit concerned about, is that there is no separate color for suspended tabs. These are by default `#000`, but with this theme any tab is, so a different color for suspended ones would be appreciated, though I'm not sure which one would look nice. (oh, and just to make sure, if you now update the file, it changes from just `2025` to `2025-2026` at the top 😉)

3mpee3mpee:
> Thanks for the update, this looks really great and compact now. Just one more thing I'm a bit concerned about, is that there is no separate color for suspended tabs. These are by default `#000`, but with this theme any tab is, so a different color for suspended ones would be appreciated, though I'm not sure which one would look nice. (oh, and just to make sure, if you now update the file, it changes from just `2025` to `2025-2026` at the top 😉)

You right, I checked couple of different colors, for me #111 works well, so updated with it. Let me know if you have any  concerns. And Happy New Year btw !!!

PR #613, opened 2025-11-21
feat(theme): add new light/dark variants and make default themes follow system color scheme

TheSast:
Since recently `set nativetheme=system` was revived, and with it becoming the default in 13.0.0, I thought it would make sense to also dynamically adjust Vieb's own colorscheme based on system preference.
I did a naive search and replace for `default.css` to `dark.css` I doubt it is desireable to leave it as is, but it works as far as I tested.
This is a draft mainly because I want feedback about how to handle `default.css` and `dark.css` and ask if perhaps it's best to add css importing instead of repeating the light colorscheme everywhere, I didn't make it into  an issue  because I also want it to be available to anyone who wants to try it out.

### DIFF
--- a/app/colors/dark-compact.css
+++ b/app/colors/dark-compact.css
@@ -1,0 +1,91 @@
+/*
+* Vieb - Vim Inspired Electron Browser
+* Copyright (C) 2022-2025 Jelmer van Arnhem
+*
+* This program is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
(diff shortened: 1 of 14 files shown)


Jelmerro:
I have been considering this option for a while now, and I'm glad to see there is demand for it and that you decided to implement this already. I do think this is a complex feature, because not all would want this, so there should indeed be separate light and dark themes, which you took care of nicely. On the other hand, I am not sure about the approach of replacing default with dark for internal code. Shouldn't the theme be reflected in these parts too? Usually the entire reason the default css is loaded is to apply the default colors, which are then possibly overridden by the active theme. So if we change the default theme to adapt to light and dark, I expect the interface to do so in all places, unless I'm missing some issues with this reasoning you encountered while implementing. Overall, I'm glad there's development on this, we can discuss it further here before we decide to merge this in, which will only happen when I start working on the v13 branch anyway, as this is definitely a breaking change.

Jelmerro:
Since this would only be merged as a breaking change anyway in the upcoming v13 branch, I don't mind having only an updated default.css, not a new-default. So basically: I would prefer this PR to show me the ideal end-state diff without migrations or handling of existing features, merely changing the defaults to do this, and have separate themes dark/light that don't.

PR #618, opened 2026-05-13
Add aria-roles to follow mode

SweatyCircle439:
I added a role follow mode that follows all elements with a role that is one of
- button
- menuitem
- menubutton
- link
- option

I also changed the way followed elements are clicked so it sends the proper events instead of calling click. This should in theory reduce suspicious behavior flagging, and i have tested it to also improve compatibility with some sites.


app/colors/default.css
```diff
@@ -60,6 +60,9 @@
     --follow-url-bg: #bff;
     --follow-url-border: #7dd;
     --follow-url-hover: #7dd7;
+    --follow-role-bg: #f60;
+    --follow-role-border: #d20;
+    --follow-role-hover: #d207;
```
Jelmerro: Either we should add these in proper colors to other themes, or we merge this type with an existing follow category.

app/preload/follow.js
```diff
@@ -119,13 +119,25 @@ const elementsWithMouseListeners = els => contextBridge.executeInMainWorld({
             "mouseover",
             "contextmenu"
         ]
+        const interactableRoles = [
```
Jelmerro: This can be changed to a Set.

app/preload/follow.js
```diff
@@ -119,13 +119,25 @@ const elementsWithMouseListeners = els => contextBridge.executeInMainWorld({
...
+            "option",
+        ]
         return allElements?.reduce((elementsWithListeners, el) => {
             if (clickEvents.some(t => el.hasAttribute(`on${t}`)
             // @ts-expect-error Only HTMLElements can have them as property.
                 || !!el[`on${t}`] || el.hasAttribute("jsaction"))) {
                 // @ts-expect-error Reduce types are broken in TS.
                 elementsWithListeners.push({el, "type": "onclick"})
             }
+            // @ts-expect-error If it is null includes will return false, not throw an error TS, so shut up please.
```
Jelmerro: This can be prevented by adding `?? ""` after the getAttribute, also removing the ts-expect-error line that is too long.

app/preload/follow.js
```diff
@@ -376,7 +388,20 @@ ipcRenderer.on("focus-input", async(_, follow = null) => {
...
+        clientX: rect.left + rect.width / 2,
+        clientY: rect.top + rect.height / 2,
+    }
+
+    focusEl.dispatchEvent(new MouseEvent("pointerdown", eventOptions));
+    focusEl.dispatchEvent(new MouseEvent("mousedown", eventOptions));
+    focusEl.dispatchEvent(new MouseEvent("pointerup", eventOptions));
+    focusEl.dispatchEvent(new MouseEvent("mouseup", eventOptions));
+    focusEl.dispatchEvent(new MouseEvent("click", eventOptions));
+
```
Jelmerro: This code is only used for input elements, but I think this should be done across the board for all type of follow clicks. Could you make a separate PR for this so that I can review this separately?

Jelmerro (changes requested):
Thank you for opening this PR. I like the core changes of this, and think we should aim to make follow mode able to recognize role elements.

I would like the detected elements with a role attribute to be in existing follow mode categories rather than new ones. These are almost always simple clickable elements (or rarely text input), at least the main ones we actually want to include from [this big complete list of aria roles](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles). I think there might be a few ones listed there that we should consider including in this too, like for example textbox or radio.

There's a whole bunch of eslint errors in this PR that I would like to have solved, some of which I have left a comment for, most of them can be fixed automatically.

The changes to the click I would like to have in a separate PR so I can review it separate from this.

SweatyCircle439:
I moved the improved click to a new branch, and i merged the role follow mode into input and click

app/preload/follow.js
Jelmerro: thanks

app/preload/follow.js
```diff
@@ -52,15 +52,29 @@ const clickInputs = [
     "input[type=\"image\"]",
     "input[type=\"reset\"]",
     "*[role=\"button\"]",
+    "*[role=\"gridcell\"]",
```
Jelmerro: I don't think the gridcell should be included.

app/preload/follow.js
```diff
@@ -52,15 +52,29 @@ const clickInputs = [
...
+    "*[role=\"link\"]",
```
Jelmerro: I think the link type might be best suited in other, but am not sure as of yet how that would work code-wise. Maybe best to exclude it for now, if you don't want to spend some effort detecting this elsewhere.

app/preload/follow.js
```diff
@@ -52,15 +52,29 @@ const clickInputs = [
...
+    "*[role=\"menuitem\"]",
+    "*[role=\"menuitemcheckbox\"]",
+    "*[role=\"menuitemradio\"]",
+    "*[role=\"menubutton\"]",
```
Jelmerro: Is menubutton a framework thing, or can this be removed? It's not in the aria list on MDN. Do you have a link to some documentation on this?

app/preload/follow.js
```diff
@@ -52,15 +52,29 @@ const clickInputs = [
...
+    "*[role=\"option\"]",
     "*[role=\"radio\"]",
     "*[role=\"checkbox\"]",
+    "*[role=\"scrollbar\"]",
```
Jelmerro: I don't think scrollbars should be clickable, this can be removed.

app/preload/follow.js
```diff
@@ -52,15 +52,29 @@ const clickInputs = [
...
+    "*[role=\"slider\"]",
```
Jelmerro: Like regular range-type inputs, the slider role should be part of textlikeInputs as an insert mode element.

Jelmerro (commented):
Great, this is much cleaner. Just some feedback/questions on the aria roles themselves now.

SweatyCircle439:
i have made the requested changes

app/preload/follow.js
Jelmerro: Using all roles that are prefixed with menu* is a good solution indeed, this includes custom roles like the previously included menubutton.

Jelmerro (approved):
Thanks for working on this.

PR #619, opened 2026-05-13
Feature/improved click

SweatyCircle439:
I improved the click handling (currently only for inputs, however this same logic could easily be applied to buttons, i just couldn't find any other refrences to HTMLElement.click() and neither could my IDE which have to be replaced by this logic).

this change works by creating and sending actual events instead of emulating them which improves compatibility and should reduce flagging for suspicious behavior on certain sites.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ node_modules
 *.swp
 logo-wip-files
 ViebData
+.idea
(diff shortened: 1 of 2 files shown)


Jelmerro (commented):
Thanks for making this PR.

Instead of only applying these changes for input elements, could we make a function for this and use it both here and in the simulated click event? That would solve it for all type of follow mode clicks. I believe a similar thing exists at the end of app/index.js already, but these are currently a bit all over the place. As such it is bit hard to tell if this is a change we should make here or elsewhere, and also what exact issue this solves for input elements, as that's not the one that has the occasional focus/blur issues, that would be click elements that are handled via simulated clicks.

I'll give this some thought and might review it properly later, but for now I need to get a clearer picture of the current system again as it's been quite a while since I worked on this part of the input handling.

PR #626, opened 2026-07-31
Feat/nix flake

0fflineuser:
Hi,

I wanted to try Vieb on nixos but the package seems to have been removed from nixpkgs.
This adds a self-contained flake so it can be built and run with Nix:
```
nix build .#vieb
nix run .#vieb
```
Also usable as a flake input from other repos.

I also added some automation (Some feedback and tests are needed, I haven't used these actions before):
- .github/dependabot.yml for weekly PRs bump to flake.lock (Not sure if weekly is to short/long for you)
- .github/workflows/nix.yml to automatically update the npmDepsHash when any of the files in 'paths' change in commits or pull_requests (package.json, package-lock.json, flake.nix, nix/**, .github/workflows/nix.yml)

I am not sure if you have any preferences for the automation or if it's good like that.

Thanks !


TheSast:
After the package got removed from nixpkgs it's maintainer made a flake for Vieb over at [tejing1/vieb-nix](https://github.com/tejing1/vieb-nix) and made the package available in the [NUR](https://github.com/nix-community/NUR).

0fflineuser:
@TheSast, thanks I didn't know!

